PR #17, opened 2016-02-11
webpack dependency needed to build

lalala:
You need webpack as a dependency to actually build the project. Unless you have it installed globally, which you shouldn't.


GU5TAF:
Nice catch, must've disappeared during some refactoring but still been in my node_modules.

Issue #5, opened 2016-02-02
Checkbox </3 redux-form


JohnPhoto:
The `<Checkbox/>` does not currently work with Redux-Form, I think the problem is the custom onChangeHandler in checkbox somehow. Will look into it and define what I'm exactly missing.

JohnPhoto:
This is the object I get from redux-form which will autowire stuff if it is `spread` to the actual `<input/>`

	{
		active: false
		defaultChecked: false
		defaultValue: undefined
		dirty: false
		error: "Required"
		initialValue: undefined
		invalid: true
		name: "person.politicalPerson"
		onBlur: function(event)
		onChange: function(event)
		onDragStart: function(event)
		onDrop: function(event)
		onFocus: function ()
		onUpdate: function(event)
		pristine: true
		touched: false
		valid: false
		visited: false
	}

Normal `<Input>` works flawless with the following code, where `name` is the same structure as object above:

    <Input type="text" label="name" { ...name }/>

JohnPhoto:
I have verified my suspicion, if I remove the "custom" `onChange` from the UIKit, the redux-logic works fine.

GU5TAF:
I think https://github.com/nordnet/nordnet-ui-kit/commit/569dfc76eae9016de7f2480479c63bef18d9a3a0 solves this.

GU5TAF:
could you perhaps confirm @JohnPhoto ?

JohnPhoto:
I think this works now, will reopen if not so.

PR #28, opened 2016-03-17
fix for redux-form compatibility

lalala:
Checkboxes aren't working properly with redux-form.

Also, the babel configuration doesn't support the destructuring spread operator, thus the ugly syntax.

### DIFF
--- a/src/components/checkbox/checkbox.jsx
+++ b/src/components/checkbox/checkbox.jsx
@@ -23,15 +23,19 @@ class Checkbox extends PureComponent {
   }
 
   renderCheckbox() {
+    const { value } = this.props;
+    const rest = this.props;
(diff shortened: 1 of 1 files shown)


GU5TAF:
could we perhaps looks at adding that to the babel config instead or using [lodash.omit](https://lodash.com/docs#omit) @lalala ?

lalala:
Sure, but I don't know what's missing :-)

lalala:
Some issues remain, need to look at how redux form deals with props, etc.

GU5TAF:
Look at #5, it might very well be the name prop that's messing it up.

GU5TAF:
Didn't solve the issue, I'm looking into it myself in the refactor/input branch.

PR #42, opened 2016-04-26
Refactored input

GU5TAF:
Cleaned up code, added some features and restructured markup.
Moved checkbox and radio into the input component.


GU5TAF:
couldn't add @johnphoto as reviewer for some reason 💩
ping @dimchez

src/components/collapsible/collapsible.jsx
```diff
@@ -76,7 +78,7 @@ class Collapsible extends PureComponent {
 
   renderContent() {
     return (
-      <div ref={ content => this.content = content }>
+      <div ref={ content => { this.content = content; } }>
```
dimchez: I'd put definition of this ref callback outside of render() call.

src/components/date-picker/index.js
```diff
@@ -0,0 +1,2 @@
+import DatePicker from './date-picker';
+export default DatePicker;
```
dimchez: In `index.js` files I prefer to use 
```
export {
  DatePicker,
};
```

It's easier to extend with additional exports later.

src/components/dropdown/dropdown.jsx
```diff
@@ -29,11 +29,11 @@ class Dropdown extends PureComponent {
 
     return (
       <ul className="dropdown__actions" style={ style }>
-        {this.props.actions.map((action, index) => (
+        { this.props.actions.map((action, index) => (
           <li className="dropdown__action" key={ index } onClick={ action.action }>
```
dimchez: I'd suggest using `${index}_${action.label}` as `key`. Simply using `index` might not be unique and consistent enough.

src/components/icon/icon.jsx
```diff
@@ -1,3 +1,4 @@
+/* eslint global-require: "warn" */
```
dimchez: I'd suggest either 
* disabling/setting to warn level this rule for one single line in this file or
* setting it to warn level globally in `eslintrc`

src/components/collapsible/collapsible.jsx
GU5TAF: the reason for this is new linter settings, particularly [no-return-assign](http://eslint.org/docs/rules/no-return-assign).

src/components/input/help-text.jsx
```diff
@@ -0,0 +1,27 @@
...
+    'help-text--has-success': hasSuccess,
+    'help-text--has-warning': hasWarning,
+    'help-text--has-error': hasError,
+  });
+
+  if (props.children) {
+    return <span className={ classes }>{ props.children }</span>;
+  }
+
+  return <span />; // Stateless functions do not support returning null, see https://github.com/facebook/react/issues/5355
```
dimchez: React 15 supports it. But I see from package.json that we declare support for v14 as well. I guess when we bump version to 1.x we should drop React v14 support.

src/components/input/input-checkbox-radio.jsx
```diff
@@ -23,66 +25,69 @@ class Checkbox extends PureComponent {
...
-          { this.state.checked && this.state.disabled ? <Icon type="checkmark" stroke="#AAAAAA" style={{ display: 'block' }} /> : null }
-        </span>
+        { this.props.type === 'checkbox' ? <Checkbox { ...this.state } /> : null }
+        { this.props.type === 'radio' ? <Radio { ...this.state } /> : null }
       </div>
     );
   }
 
   renderLabel() {
     if (!this.props.label) return null;
```
dimchez: Please, don't use one liners like this!

src/components/input/help-text.jsx
GU5TAF: I know, should probably change the line comment to `Stateless functions in React 0.14.X do not support returning null, see https://github.com/facebook/react/issues/5355`

src/components/input/input-default.jsx
```diff
@@ -0,0 +1,196 @@
...
+      hasValue: this.hasValue(event.target.value),
+    });
+
+    if (this.props.onChange) {
+      this.props.onChange(event);
+    }
+  }
+
+  hasValue(value) {
+    return isNumber(value) || (value && !!value.length);
```
dimchez: Having `(value && value.length)` would yield same result. Both options work though :)

src/components/input/input-password.jsx
```diff
@@ -0,0 +1,73 @@
...
+    this.state = Object.assign({}, this.state, {
+      showPassword: !!props.showPassword,
+      hasAddon: true,
+    });
+
+    this.renderPasswordToggle = this.renderPasswordToggle.bind(this);
+    this.onPasswordToggleChange = this.onPasswordToggleChange.bind(this);
+  }
+
+  // TODO: How to handle this when there's already a componentWillReceiveProps method?
```
dimchez: You can do `super.componentWillReceiveProps()` if you need to call parents class method.

src/components/input/input-select.jsx
```diff
@@ -0,0 +1,59 @@
...
+        <select
+          { ...this.props }
+          id={ this.props.id }
+          className={ classes }
+          onFocus={ this.onFocus }
+          onBlur={ this.onBlur }
+          onChange={ this.onChange }
+          placeholder=""
+          value={ this.state.value }
+          style={ this.state.hasFocus || this.state.value.length > 0 ? { } : { color: 'transparent' } }
```
dimchez: No need to compare with zero, just `this.state.value.length` is enough.

src/components/input/input-select.jsx
```diff
@@ -0,0 +1,59 @@
...
+        >
+          <option value="" disabled>{ this.props.placeholder }</option>
+          { this.props.options.map(option => <option key={ option.value } value={ option.value }>{ option.label }</option>) }
```
dimchez: I'd use `${index}_${option.value}` for `key`.

src/components/pane/pane.jsx
```diff
@@ -23,28 +23,28 @@ class Pane extends PureComponent {
   renderTabs() {
     return (
       <ul className="pane__tabs">
-        {this.props.tabs.map((tab, index) => (
+        { this.props.tabs.map((tab, index) => (
           <li
             key={ index }
```
dimchez: I'd suggest using `${index}_${tab.label}` for `key`.

src/components/pane/pane.jsx
```diff
@@ -23,28 +23,28 @@ class Pane extends PureComponent {
...
-            className={ classNames('pane__tab', { 'pane__tab--is-active': this.state.activeTab === index })}
-            style={{ width: `${100 / this.props.tabs.length}%` }}
+            className={ classNames('pane__tab', { 'pane__tab--is-active': this.state.activeTab === index }) }
+            style={ { width: `${100 / this.props.tabs.length}%` } }
```
dimchez: I think this should be moved outside of `map` function since it can be calculated once instead of doing same calculation on every iteration.

src/components/pane/pane.jsx
```diff
@@ -23,28 +23,28 @@ class Pane extends PureComponent {
...
     );
   }
 
   renderBody() {
     return (
       <div className="pane__body">
-        {this.props.tabs.map((tab, index) => (
-          <div key={ index } style={{ display: this.state.activeTab === index ? 'block' : 'none' }}>
+        { this.props.tabs.map((tab, index) => (
+          <div key={ index } style={ { display: this.state.activeTab === index ? 'block' : 'none' } }>
```
dimchez: Consider using `body_${index}_${tab.label}` for `key`.

test/components/dropdown/dropdown.test.js
```diff
@@ -6,9 +6,13 @@ import Dropdown from '../../../src/components/dropdown/dropdown';
...
   let wrapper;
 
-  beforeEach(() => wrapper = shallow(
-    <Dropdown toggle="Dropdown" actions={[{ label: 'log', action() {console.log('I was clicked!');} }]} />
-  ));
+  beforeEach(() => {
+    wrapper = shallow(
+      <Dropdown
+        toggle="Dropdown"
+        actions={ [{ label: 'log', action() { console.log('I was clicked!'); } }] }
```
dimchez: Don't use `console.log()` in tests. It's an external dependency that should not be present in unit tests. Either supply a spy if you need to verify that action was triggered or use empty function.

src/components/date-picker/index.js
dimchez: Keep it as is though since `webpack.config.babel.js` would blow up if changing this export.

src/components/input/help-text.jsx
dimchez: Yeah, changing it like that would be good.

GU5TAF:
Thanks for all the feedback @dimchez, will look at solving these then I'll ping @johnphoto
dmitry

JohnPhoto:
Remember to guard the select-input when options is missing:

```
/Users/johbac/ws/git/order/node_modules/nordnet-ui-kit/dist/input/index.js:279
	          this.props.options.map(function (option) {
	                            ^

TypeError: Cannot read property 'map' of undefined
```

JohnPhoto:
On the same note, please make the select look nice when no options (empty array) is given.

GU5TAF:
I think this bad boy is ready to be merged @JohnPhoto @dimchez, thoughts?
click it

JohnPhoto:
Looks good to me, but got no power here.

PR #44, opened 2016-04-27
Updated value a bit

JohnPhoto:
It now accepts string, number, element for both label and value.


GU5TAF:
I've already added these features amongst some others locally so will decline this PR in favour of that one.

PR #49, opened 2016-06-07
Feature/spark graph responsive

robineng:
Added a responsive spark graph component that redraws to keep stroke width if height and width aren't set.


GU5TAF:
awesome work @robineng 🎉

Issue #60, opened 2016-07-15
Could we get a spinner component?


macber:
A default spinner with height and width and whatnot....

robineng:
See #61 for pull request.

robineng:
Merged

PR #61, opened 2016-07-19
Feature/spinner component

robineng:
Added spinner component as requested in #60


package.json
```diff
@@ -51,7 +51,8 @@
     "react-prop-types": "^0.3.0",
     "react-pure-render": "^1.0.2",
     "react-select": "^1.0.0-beta13",
-    "react-svg-sprite-icon": "^0.0.11"
+    "react-svg-sprite-icon": "^0.0.11",
+    "spinkit": "^1.2.5"
```
GU5TAF: It looks like this dependency isn't used, right?

robineng: It is used in the spinner.scss

src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,8 @@
+@import '../../../node_modules/spinkit/scss/spinners/9-cube-grid',
```
GU5TAF: this can be written as `@import '~spinkit/scss/spinners/9-cube-grid'` as web pack resolves ~ to node_modules

src/components/spinner/spinner.scss
```diff
@@ -0,0 +1,8 @@
...
+        '../../_variables';
```
GU5TAF: please declare all imports on their own line

package.json
GU5TAF: Ah, I see that now. My bad 😁

PR #85, opened 2016-09-30
Don't mess with the Z-Index

jesperwiner:


iamstarkov:
Can you elaborate what's the problem with this?

jesperwiner:
There is a z index conflict between range selector and search result

Cssnano tries to be smart during scss to CSS compilation but fails

Gustaf said to just turn it off

PR #88, opened 2016-10-07
Search component fixes

jesperwiner:


src/components/search/index.js
```diff
@@ -1 +1,2 @@
-export { default } from './search';
+import Search from './search-container';
+export default Search;
```
GU5TAF: You can use the same 1 line pattern but with the search container instead.

src/components/search/search-container.jsx
```diff
@@ -0,0 +1,4 @@
+import Search from './search';
+import onOutsideClick from 'react-onclickoutside';
+
+export default onOutsideClick(Search);
```
GU5TAF: Please just call it `onClickOutside`, swapping it around hurts my brain.

PR #106, opened 2016-12-13
removed outline glow when active

friendlyduck:
On:
- buttons
- dropdown
- range selector

### DIFF
--- a/src/components/dropdown/dropdown.scss
+++ b/src/components/dropdown/dropdown.scss
@@ -15,6 +15,7 @@
     z-index: 4;
     font-weight: 600;
     font-size: rem(14px);
+    outline: none;
 
     &-icon {
       width: rem(10px);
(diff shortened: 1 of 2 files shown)


src/components/dropdown/dropdown.scss
```diff
@@ -15,6 +15,7 @@
     z-index: 4;
     font-weight: 600;
     font-size: rem(14px);
+    outline: none;
```
iamstarkov: why?

friendlyduck: @iamstarkov What if we remove it from the dropdown input (as this is not implemented consistently on other inputs), and keep it for the range selector, but with a blue background? Like the attached image? 
image

iamstarkov:
close this PR as far as we are completely reworking ui kit

PR #107, opened 2016-12-21
Fixing support for node 4

bstream:


bstream:
When running `npm start` without webpack-hot-middleware installed, webpack seems to be confused about how to find it, see image below:

image

That error was solved when I npm installed webpack-hot-middleware.

bstream:
@iamstarkov Merging this since some people need it now. Lets discuss it when you come back :)

PR #108, opened 2017-01-10
5Feature/tooltip

KitsuneKelso:
@bstream Using a slightly improved and more stable version of the tooltip I developed for the OP projects. When time allows I motion to try to incorporate some of the more customizable options we saw in the other lib we were looking at.

### DIFF
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -43,13 +43,13 @@ class Tooltip extends React.Component {
     });
   }
 
-  renderPopup(text) {
+  renderPopup(content) {
     if (this.state.hover || this.state.toggled) {
       return (
         <div className="tooltip-popup" onMouseEnter={ this.mouseEnter } onMouseLeave={ this.mouseLeave }>
           <div className="tooltip-popup__content">
             <div className="tooltip-popup__content-text">
-              { text }
+              { content }
             </div>
           </div>
         </div>
@@ -58,20 +58,20 @@ class Tooltip extends React.Component {
     return null;
   }
 
-  renderTooltip(text) {
+  renderTooltip(content) {
     return (
       <div className="tooltip">
         <div className="tooltip-container" onClick={ this.toggleShow } onMouseEnter={ this.mouseEnter } onMouseLeave={ this.mouseLeave }>
           { this.props.container }
         </div>
-        { this.renderPopup(text) }
+        { this.renderPopup(content) }
       </div>
     );
   }
 
   render() {
     return (
-      this.renderTooltip(this.props.text)
+      this.renderTooltip(this.props.content)
     );
   }
 }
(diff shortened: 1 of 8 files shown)


src/components/tooltip/tooltip.scss
```diff
@@ -0,0 +1,93 @@
+$z-index-tooltip: 2;
+$z-index-tooltip-closer: 5;
+
+.tooltip {
+  color: #00A9EC;
+  margin-left: 10px;
```
bstream: Use `rem`-function from variables. See other scss files. Usage is `rem(10px)`

src/components/tooltip/tooltip.scss
```diff
@@ -0,0 +1,93 @@
...
+  p {
+    margin: 0;
+  }
+  &-questionmark {
+    width: 16px;
+    height: 16px;
+    cursor: pointer;
+    display: inline;
+    &:focus {
+      outline:0;
```
bstream: Also hurting accessibility. Better with a custom outline if the original is ugly 😎

src/components/tooltip/tooltip.scss
```diff
@@ -0,0 +1,93 @@
...
+      top: 0;
+      right: 0;
+      bottom: 0;
+      left: 0;
+      z-index: $z-index-tooltip-closer;
+    }
+  }
+}
+
+@media only screen and (max-width: 62em) {
```
bstream: Maybe drop media queries in the lib and then the users can add them themselves so that they can have other breakpoints etc.

KitsuneKelso: Sounds like a good idea, will do!

src/components/tooltip/tooltip.scss
KitsuneKelso: You're absolutely correct. I think we removed the outline on some request previously and this has lived on in the .scss file since then. Will fix!

src/components/tooltip/tooltip.jsx
```diff
@@ -0,0 +1,70 @@
...
+        <div className="tooltip-popup__closer" onClick={ this.toggleShow }></div>
+      );
+    }
+    return null;
+  }
+
+  renderTooltip(text) {
+    return (
+      <div className="tooltip">
+        <div className="tooltip-questionmark" tabIndex="-1" onClick={ this.toggleShow }>
```
KitsuneKelso: So essentially remove tabindex and allow for hover functionality on desktop. Gotcha! :+1:

src/components/tooltip/tooltip.scss
KitsuneKelso: Will have a look!

KitsuneKelso:
@bstream I have updated the PR with fixes according to your comments. The only thing currently still not working as intended is the toggle feature as I have not gotten the react-onclickoutside lib working. Will still try to fix this, but using the closer div until then to still allow for closing on mobile safari.

PR #111, opened 2017-01-17
Feature/tooltip changes

bstream:
Made some changes to the `<Tooltip />` component.

* Less default styles.
* Option to pass in custom `className`.
* `container` -> `children`.

### DIFF
--- a/src/components/tooltip/tooltip.scss
+++ b/src/components/tooltip/tooltip.scss
@@ -4,38 +4,33 @@
 $z-index-tooltip: 2;
 
 .tooltip {
-  color: #00A9EC;
-  margin-left: rem(10px);
+  color: $color-primary;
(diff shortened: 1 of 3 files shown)


src/components/tooltip/tooltip.scss
```diff
@@ -46,10 +41,11 @@ $z-index-tooltip: 2;
       width: 0;
       height: 0;
       border-style: solid;
-      border-width: 0 rem(6px) rem(8px) rem(6px);
+      border-width: 0 rem(8px) rem(8px) rem(8px);
       border-color: transparent transparent rgba(0, 0, 0, 0.75);
       z-index: $z-index-tooltip;
     }
+
     &:after {
```
KitsuneKelso: This after can be removed. ;)

PR #116, opened 2017-02-20
Removed react-date-picker and associated code. If we need a styled da…

jesperwiner:
Removed react-date-picker and associated code. If we need a styled datepicker later we can make a new repo for it.

Were so many issues trying to have it external with we using it inside the input type date as well, 
and the only place we use a date picker we don't even use this one...


iamstarkov:
Can i ask you to create initial changelog with this breaking change being mentioned there?

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Changelog
 
+## 0.0.58
```
iamstarkov: usually at the time of pull-request being submitted you dont know in what version it will be released, so rule of thumb is to keep it simply `## upcoming version`

jesperwiner: ok

iamstarkov:
and my concern is that you want to release breaking change as a patch. yeah, we are on 0.x.x track. but nn-ui-kit wouldnt like breaking change as patch change. lets do it in minor bump

iamstarkov:
approved, but please dont release it as patch

iamstarkov:
@jesperwiner "delete branch" ping

PR #117, opened 2017-02-24
Webpack 2

robineng:
* Updated to Webpack 2
* Bumped styleguidist
* Removed some unused dependencies
* Replaced lodash.assign with Object.assign
* Replaced react-pure-render with React.PureComponent
* Replaced a few other lodash modules with own code.


src/utilities/omit.js
```diff
@@ -0,0 +1,5 @@
+export default function omit(object = {}, ...keys) {
+  const copy = Object.assign({}, object);
+  keys.forEach(key => delete copy[key]);
+  return copy;
+}
```
iamstarkov: its not optimal

iamstarkov: delete is direct deoptimization for v8

iamstarkov: as far as you do copy on your own. its better to create fresh copy and then add every non-omitted key-val into copy from original object

iamstarkov: for the record `Object.assign` is shallow operation, so its not true copy. but its okay in this case

robineng: > delete is direct deoptimization for v8

Please explain :D

> as far as you do copy on your own. its better to create fresh copy and then add every non-omitted key-val into copy from original object.

That would the time complexity O(objSize * keysLength) instead of O(objSize + keysLength) assuming that delete is constant but my assumption may be wrong due to your prevoius comment that I do not fully grasp

> for the record Object.assign is shallow operation, so its not true copy. but its okay in this case

Yes, might have been a poor choice of variable naming. My intention was not to do a deep copy.

iamstarkov: > ### 5.2.1. Objects that are in "hash table mode" (aka "normalized objects", "dictionary mode" - objects who have a hash table as a backing data structure) are not simple enumerables**
> 
> ```js
> function hashTableIteration() {
>     var hashTable = {"-": 3};
>     for(var key in hashTable);
> }
> ```
>
> An object will go into hash table mode for example when you add too many properties dynamically (outside constructor), delete properties, use properties that cannot be valid identifiers and so on. In other words, when you use an object as if it was a hash table, it will be turned into a hash table. Passing such an object to for-in is a no no. You can tell if an object is in hash table mode by calling console.log(%HasFastProperties(obj)) when the flag --allow-natives-syntax is enabled in Node.JS.
> — ["Optimization killers"](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#521-objects-that-are-in-hash-table-mode-aka-normalized-objects-dictionary-mode---objects-who-have-a-hash-table-as-a-backing-data-structure-are-not-simple-enumerables) by petkaantonov (bluebird's author, and bluebird is fastest promises implementation)

iamstarkov: > That would the time complexity O(objSize * keysLength) instead of O(objSize + keysLength)

you are right, but your proposal is not possible to optimise. Mine will be warmed and jit-ed. So my proposal will be slower only first time, i would say

iamstarkov: btw, why did you decide to go with in-house implementation instead of lodash's one?

robineng: Cool, changed it now.

> btw, why did you decide to go with in-house implementation instead of lodash's one

lodash.omit was like 30kB unminified, I cant remember exact minified stats. But I felt that for the simple use cases we had we didn't need it.

package.json
```diff
@@ -43,83 +44,66 @@
   "dependencies": {
     "array-equal": "^1.0.0",
     "bezier-easing": "^2.0.3",
-    "chroma-js": "^1.1.1",
     "classnames": "^2.2.3",
     "color": "^0.11.3",
     "d3-interpolate": "^1.0.0",
-    "date-arithmetic": "^3.1.0",
-    "lodash.assign": "^4.0.9",
     "lodash.camelcase": "^4.1.0",
```
bstream: We might want to combine all these `lodash.*` imports now when we have webpack 2.

iamstarkov: btw, are you aware about lodash's [webpack](https://www.npmjs.com/package/lodash-webpack-plugin) and [babel](https://www.npmjs.com/package/babel-plugin-lodash) plugins?

these can help even further

src/utilities/omit.js
```diff
@@ -1,5 +1,6 @@
 export default function omit(object = {}, ...keys) {
-  const copy = Object.assign({}, object);
-  keys.forEach(key => delete copy[key]);
-  return copy;
+  return Object.keys(object)
+    .filter(objectKey => keys.every(omitKey => objectKey !== omitKey))
+    .map(key => ({ [key]: object[key] }))
+    .reduce((prev, entry) => Object.assign(prev, entry), {});
```
iamstarkov: thats beautiful

package.json
robineng: I can do some experimenting when I'm back from vacation 😀

iamstarkov: @robineng  have a fun vacation!

robineng: I've been doing some testing with the lodash webpack and babel plugins. It has both pros and cons. 

Pros: 
Neater package.json (less deps), less import statements for components that uses multiple lodash functions, -3kB bundle size

Cons: 
When altering the set of used lodash functions we have to manually maintain the lodash-webpack-plugin options object by removing or adding features if needed.

I'm updating the PR

robineng:
Copying a comment I wrote on outdated code a few seconds ago: 

I've been doing some testing with the lodash webpack and babel plugins. It has both pros and cons.

Pros:
Neater package.json (less deps), less import statements for components that uses multiple lodash functions, -3kB bundle size

Cons:
When altering the set of used lodash functions we have to manually maintain the lodash-webpack-plugin options object by removing or adding features if needed.

The PR is updated

PR #118, opened 2017-03-20
New component TextIcon

johankitti:


src/components/text-icon/text-icon.scss
```diff
@@ -0,0 +1,8 @@
+.icon {
+  display: flex;
+  justify-content: center;
+  align-items: center;
+  height: 5rem;
```
bstream: Use `rem` function instead. See examples in other scss files

src/components/text-icon/text-icon.jsx
```diff
@@ -0,0 +1,29 @@
...
+}
+
+TextIcon.propTypes = {
+  text: PropTypes.string.isRequired,
+  color: PropTypes.string,
+  textColor: PropTypes.string,
+};
+
+TextIcon.defaultProps = {
+  color: '#4C4A56',
```
bstream: Use variables for colors, also set them in the class for now (instead of style)

PR #127, opened 2017-04-05
Quickfix making table components not rely on scss

willeeklund:
Quickfix making table components not rely on scss.

I have added `TODO: Move SCSS into JSS` to make us remember to move the styling as well.


iamstarkov:
@bstream pls review, and merge if its fine by you

PR #133, opened 2017-04-07
Fast/more components

bstream:
Fast port: Search, SparkGraph, Spinner, Widget

Fix styleguidist warnings in a few other components

### DIFF
--- a/src/components/input/validation-icon.jsx
+++ b/src/components/input/validation-icon.jsx
@@ -1,23 +1,23 @@
 import React from 'react';
-import Icon from '../icon/icon';
+import { Icon } from '../../';
 // import variables from '../../utilities/variables';
 
 function ValidationIcon(props) {
   const iconClass = 'input__validation-icon';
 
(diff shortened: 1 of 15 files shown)


src/components/input/validation-icon.jsx
```diff
@@ -1,23 +1,23 @@
...
+    return <Icon.ExclamationPoint className={iconClass} stroke={'variables.colorWarning'} />;
   }
 
   if (props.hasError) {
-    return <Icon className={iconClass} type="exclamationPoint" />;
+    return <Icon.ExclamationPoint className={iconClass} />;
   }
 
-  return <span />; // Stateless functions in React 0.14.X do not support returning null
+  return null; // Stateless functions in React 0.14.X do not support returning null
```
willeeklund: If returning null is now supported, I think we should remove the comment on this line.

bstream: Will put that in the next PR #134

PR #137, opened 2017-04-11
Transform Dropdown component to use JSS

willeeklund:
screen shot 2017-04-11 at 15 01 30


src/components/dropdown/dropdown-styles.jsx
```diff
@@ -0,0 +1,50 @@
...
+export default createStyleSheet('Dropdown', (theme) => {
+  const { palette } = theme;
+
+  return {
+    dropdown: {
+      display: 'inline-block',
+      position: 'relative',
+      marginBottom: '16px',
+
+      '& .dropdown__toggle': {
```
bstream: Remove bem and flatten object

willeeklund: Fixed now.

PR #142, opened 2017-04-12
Search component with JSS

willeeklund:
screen shot 2017-04-12 at 17 05 54


src/components/search/search-styles.jsx
```diff
@@ -0,0 +1,163 @@
...
+      fontWeight: 600,
+      marginBottom: '4px',
+      whiteSpace: 'nowrap',
+    },
+
+    market: {
+      fontSize: '12px',
+      color: palette.text.default,
+
+      '& .flag': {
```
bstream: Does `.flag` here really work? Rather send in a custom className to flag and style it here

willeeklund: Good catch! I added className="flag" to the flag component to be able to style it.

PR #145, opened 2017-04-13
Table components with JSS

willeeklund:
screen shot 2017-04-13 at 15 46 53


src/components/td/td.jsx
```diff
@@ -20,38 +19,33 @@ function Td({
...
-    'td--align-center': align === 'center',
-    'td--ellipsis': ellipsis,
-  }, className);
+}, { styleManager }) {
+  const classes = styleManager.render(TdStyles);
+  const usedClassName = classNames(
+    classes.td, size,
+    modifier ? [modifier] : [],
+    highlight ? [`highlight-${highlight}`] : [],
+    align ? [`align-${align}`] : [],
```
bstream: These lines can be expressed like this:

```js
{
  [`${modifier}`]: modifier,
  [`highlight-${highlight`]: highlight,
  [`align-${align}`]: align,
}
```

src/components/table/style-utilities.jsx
```diff
@@ -0,0 +1,132 @@
+import c from 'color';
```
bstream: There are a lot of classes left in this file that does not utilize jss (`xs`/`sm`/..., `mono`, `ellipsis`, etc). Is that necessary to make the table styling work?

src/components/td/td.jsx
willeeklund: I changed to this code style now.

PR #149, opened 2017-04-18
Select using JSS but not importing css from react-select

willeeklund:
This is not completely done because I do not know how we want to include the css lib files for `react-select`. To make the styleguide work I have temporarily added it as a style-tag in `select.md`.

Apart from that it has been ported.

screen shot 2017-04-18 at 15 02 24


bstream:
Lets remove this component since it is not used

PR #162, opened 2017-04-21
Added new avatar size after request from team

willeeklund:
Added new avatar size "xs" after request from team.

screen shot 2017-04-21 at 10 07 18


willeeklund:
Added it to example file and merged to master.

PR #168, opened 2017-04-24
Added option on segmented control to work as a toggle button

willeeklund:
screen shot 2017-04-24 at 16 59 45

New API where you specify `type` (`radio` or `checkbox`).
With the `checkbox` type, you can add `checked` prop to set start value. See markdown example.

### DIFF
--- a/src/components/segmented-control/segmented-control.jsx
+++ b/src/components/segmented-control/segmented-control.jsx
@@ -5,8 +5,9 @@ import SegmentedControlStyles from './segmented-control-styles';
 class SegmentedControl extends React.PureComponent {
   constructor(props, context) {
     super(props, context);
-    this.state = { selected: props.value };
+    this.state = { selected: props.value, toggleValue: props.toggleValue };
     this.handleChange = this.handleChange.bind(this);
+    this.toggleStateValue = this.toggleStateValue.bind(this);
     this.classes = this.context.styleManager.render(SegmentedControlStyles);
   }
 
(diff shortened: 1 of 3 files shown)


iamstarkov:
lets add examples with icons also

src/components/segmented-control/segmented-control.jsx
```diff
@@ -5,46 +5,107 @@ import SegmentedControlStyles from './segmented-control-styles';
...
-      if (this.props.onChange) {
-        this.props.onChange(index);
-      }
+    if (this.props.type === 'radio') {
+      this.setState({
+        selected: index,
+      }, () => {
+        if (this.props.onChange) {
+          this.props.onChange(this.childSelectedState());
+        }
```
bstream:
```
if (this.props.onChange) {
    this.props.onChange(this.childSelectedState());
}
```
This can be moved down since it is the same for both cases

willeeklund: Good catch. I moved it into local function that is called.

willeeklund:
I have added icons in the example as well now.

src/components/segmented-control/segmented-control.jsx
```diff
@@ -5,46 +5,104 @@ import SegmentedControlStyles from './segmented-control-styles';
...
   constructor(props, context) {
     super(props, context);
-    this.state = { selected: props.value };
+    if (this.props.type === 'radio') {
+      this.state = {
+        selected: props.value,
+      };
+    } else if (this.props.type === 'checkbox') {
+      // If multiple children
+      if (this.props.children.reduce) {
```
bstream: Can do `Array.isArray(this.props.children)` if you want to make sure this is an array :)

PR #178, opened 2017-04-25
Tweaking

bstream:
* Tweaking inputs further with addons (added new examples for these)
* Now gets the correct fonts
* Remove remaining `rem` since they were based on html fontSize being 62.5%.


src/components/input/input-default-styles.jsx
```diff
@@ -121,23 +122,27 @@ export default createStyleSheet('InputDefault', (theme) => {
...
+            display: 'flex',
+            alignItems: 'baseline',
           },
 
           '&__element': {
             width: '100%',
           },
 
           '&__addon': {
-            position: 'absolute',
```
willeeklund: I guess this removed line in intentional.

bstream: yes it is :)

PR #182, opened 2017-04-26
Allow uppercase country codes on flag

JCB-K:
only that's really changed is line 49 and onwards, rest is just formatting


JCB-K:
@bstream I've reverted the refactoring.
@iamstarkov updated CHANGELOG.md

PR #184, opened 2017-04-27
Tweaking style on Tooltip component

bstream:


JCB-K:
I ran prettier on it so if it's inconsistent, blame them 😉. But I can
revert the refactor and keep the added prop types.

PR #188, opened 2017-04-28
Updated Flag component with new combined SVG flags

willeeklund:
screen shot 2017-04-28 at 21 37 51

This implementation has been tested successfully on latest Chrome, Firefox and Safari (Mac versions).


src/components/flag/currencyFlags.jsx
```diff
@@ -0,0 +1,15 @@
+import euDk from './../../flags/currencies/eu-dk';
+import euNo from './../../flags/currencies/eu-no';
+import euSe from './../../flags/currencies/eu-se';
+import dkSe from './../../flags/currencies/dk-se';
+import noSe from './../../flags/currencies/no-se';
+import usSe from './../../flags/currencies/us-se';
```
bstream: We might want to look into using require instead of import here to reduce the size of this file in the future

PR #196, opened 2017-05-04
Added additional className props for labeled-value component.

johankitti:


bstream:
The aim of this PR was to be able to override the style on one specific size of LabeledValue, since some values in ui-kit did not correspond with the design. We discussed adding the correct size to ui-kit instead to keep it consistent.

PR #200, opened 2017-05-08
Feature/dynamic currencies

bstream:
image


src/components/flag/flag.jsx
```diff
@@ -53,17 +54,17 @@ Flag.defaultProps = {
...
-
-const countryCodes = dupInUpperCase(['ca', 'de', 'fr', 'ru', 'gb', 'dk', 'fi', 'no', 'se', 'us', 'jp', 'cn', 'eu']);
-const currencyCountryCodes = dupInUpperCase(['dk', 'no', 'se', 'us', 'eu']);
-
 Flag.propTypes = {
   className: PropTypes.string,
   /** A valid 2-character country code */
-  countryCode: PropTypes.oneOf(countryCodes),
-  secondaryCountryCode: PropTypes.oneOf(currencyCountryCodes),
+  countryCode: PropTypes.oneOf([
```
JCB-K: at least make this array a variable?

src/components/flag/flags/eu.jsx
```diff
@@ -0,0 +1,40 @@
+import React from 'react';
+
+export default function FlagRu({
```
JCB-K: Should be `FlagEu`, no?

src/components/flag/flag.jsx
iamstarkov: i would like that as well, problem is in a tooling. styleguidist doesnt evaluate variables, so we have to inline it

bstream: I did this change so that the proptypes show up in styleguidist, sadly styleguidist does not work well with variables 😭

Before:
image

After:
image

JCB-K: it's an issue even further upstream. https://github.com/reactjs/react-docgen/issues/36

bstream: Hang on, it was because there was a function before I think, seems to work with a variable. Let me revise! :)

bstream:
Fixed @iamstarkov comment in eed4e7b

PR #214, opened 2017-05-12
Add new modifier to proptypes

gish:


gish:
ping @bstream @willeeklund @iamstarkov

willeeklund:
Thanks Erik! Good catch.

PR #217, opened 2017-05-15
Set pointer-events to none on icons

JCB-K:
We have some strange error where icons inside react-router <Link /> components are not clickable.
By setting pointer-events to none the event bubbles up to the parent.

Also moved some boilerplate from index.js to icon/index.js.

### DIFF
--- a/test/components/icon/icon.test.js
+++ b/test/components/icon/icon.test.js
@@ -31,4 +31,14 @@ describe('<Icon />', () => {
     wrapper = shallow(<Icon.Trash fill={color} />);
     expect(wrapper.find('g > g').prop('fill')).to.equal(color);
   });
+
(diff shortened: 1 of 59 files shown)


bstream:
Rather set this in defaultProps for icon.jsx, then it will propagate to all icons :)

JCB-K:
merge #218 before reviewing this

test/components/icon/icon.test.js
```diff
@@ -22,6 +22,7 @@ describe('<Icon />', () => {
 
   it('should be able to override stroke', () => {
     const color = 'tomato';
+    // Icon.forEach(i => console.log(i.style))
```
bstream: Ugh!

JCB-K: check again :smirk:

PR #224, opened 2017-05-17
fix(Button): Update button color 

willeeklund:
Update button color for primary and link case to be more consistent with secondary.

Before:
screen shot 2017-05-17 at 16 58 10

After:
screen shot 2017-05-17 at 16 58 19


bstream:
Input from design is - "Let's keep it as it is for now". Closing this :)

PR #234, opened 2017-05-30
Animate enhancement

willeeklund:
Two things solved in this PR:

* Require prop "animationName" when creating an Animate. 
* Add wrapping div around children of Animate to solve the problem when the children has top or bottom padding

Side effect: Now we should only have single child in each Animate.

### DIFF
--- a/src/components/animate/animate.jsx
+++ b/src/components/animate/animate.jsx
@@ -9,17 +9,15 @@ import easings from '../../styles/transitions/easings';
 class Animate extends React.PureComponent {
   constructor(props, context) {
     super(props, context);
-    this.idNbr = Math.round(Math.random() * 1000);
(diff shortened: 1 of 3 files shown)


src/components/animate/animate.jsx
```diff
@@ -52,6 +54,7 @@ class Animate extends React.PureComponent {
 Animate.propTypes = {
   children: PropTypes.node,
   className: PropTypes.string,
+  animationName: PropTypes.string.isRequired,
```
bstream: Can you add a javadoc comment above this line to say that this needs to be set for the component to render anything. (This will be converted to a comment in styleguidist). Like so:

`/** This needs to be set in order for the component to render anything */`

willeeklund: Good idea. This has been added now.

PR #236, opened 2017-06-05
fix(Table): 💄 Make horizontal scrolling work on iOS devices

bstream:
* Change breakpoint from sm to md
* Change overflow to scroll instead of auto on iOS
* Add -webkit-overflow-scrolling: touch to make scrolling smoother on
iOS


iamstarkov:
you decide to not go with global `touch` class?

bstream:
I decided to postpone the `touch` / `no-touch`. Lets discuss it later!

PR #238, opened 2017-06-05
feat(Input): ✨ Add support for autoFocus

bstream:


bstream:
Scratch this "feature" it is already supported 😂

PR #243, opened 2017-06-12
fix(Input): Remove out-of-sync marginBottom from input field in focus

willeeklund:
This created a bug in webapp-next when changing focus on the Search input field


src/components/input/input-default-styles.jsx
```diff
@@ -70,7 +70,6 @@ export default createStyleSheet('InputDefault', theme => {
       },
 
       '&.input--has-focus': {
-        marginBottom: inputMarginBottom,
```
bstream: This will break input fields that are rendered inline (was fixed in #240).

PR #245, opened 2017-06-16
fix(Alert): Avoid conflict with other header classnames

codler:


bstream:
First: Thanks for the PR! ✨
Second: I would like to get away from using bare classnames so I will try to fix that and make a separate PR for that. If I cannot get it to work soon™ I will merge this :)

bstream:
I now have a suggestion for this in #247. Check it out and see if it fits your needs :) @codler

PR #261, opened 2017-08-24
fix(Alert): change borderLeft width from 2px to 4px

iamstarkov:


iamstarkov:
that is to release Adrian changes

PR #268, opened 2017-08-30
style(Button): update style according to design

robineng:
Increase padding and descrese font size for all button sizes

### DIFF
--- a/src/components/button/button-styles.jsx
+++ b/src/components/button/button-styles.jsx
@@ -102,27 +102,31 @@ export default theme => {
     },
 
     xs: {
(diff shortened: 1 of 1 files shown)


src/components/button/button-styles.jsx
```diff
@@ -102,27 +102,31 @@ export default theme => {
     },
 
     xs: {
-      fontSize: 12,
+      fontSize: 10,
+      fontWeight: 300,
```
bstream: Use fontWeights from typography part of the theme here

robineng:
Making new PR

PR #277, opened 2017-09-13
fix(Input): Remove marginBottom if no label is given

bstream:
Changes proposed:
* If there is no label the input will not get any bottom margin compensating for that label.

Another approach would be to have another prop like `noMarginBottom` and remove the margin then. What do you guys think?


robineng:
I think that both approaches are fine. The only problem I see is if a consumer uses several inputs in the same container and only labels some of them, forcing the consumer to add margin to the unlabelled ones. It does however feel unlikely that a consumer choses only to label some of the inputs.

The `noMarginBottom` solution is more convenient as it does not result in a breaking change.

PR #296, opened 2017-10-17
fix(ProgressBar): Layout was squeezed on small screens. Going with ro…

stigdanielek:
…w wrap instead.

Before:
image

After:
image

### DIFF
--- a/src/components/progress-bar/progress-bar-styles.jsx
+++ b/src/components/progress-bar/progress-bar-styles.jsx
@@ -57,18 +57,12 @@ const variantModifierFn = (variant, colors) => {
   };
 };
 
-const sizeModifierFn = ({ typography }, { fontSize, radius }) => ({
-  borderRadius: '50%',
-  borderWidth: fontSize < 12 ? 1 : 2,
-  fontSize,
-  fontWeight: fontSize < 12 ? typography.fontWeightLight : typography.fontWeightRegular,
-  height: 2 * radius,
-  width: 2 * radius,
-
-  '& + &': {
-    marginLeft: radius,
-  },
-});
+const sizeDict = {
+  xs: { fontSize: 10, radius: 8 },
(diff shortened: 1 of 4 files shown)


src/components/progress-bar/progress-bar-styles.jsx
```diff
@@ -84,6 +88,12 @@ export default theme => {
 
     progressBar: {
       display: 'flex',
+      flexWrap: 'wrap',
+
+      '&.xs': { margin: -sizeDict.xs.radius / 2 },
+      '&.sm': { margin: -sizeDict.sm.radius / 2 },
+      '&.md': { margin: -sizeDict.md.radius / 2 },
+      '&.lg': { margin: -sizeDict.lg.radius / 2 },
```
bstream: You could replace the four lines above with `margin: ({ size }) => -sizeDict[size].radius / 2,`. What do you think @iamstarkov?

src/components/progress-bar/progress-bar-styles.jsx
```diff
@@ -119,10 +129,10 @@ export default theme => {
...
     },
 
-    xs: sizeModifierFn(theme, { fontSize: 10, radius: 8 }),
-    sm: sizeModifierFn(theme, { fontSize: 11, radius: 16 }),
-    md: sizeModifierFn(theme, { fontSize: 14, radius: 20 }),
-    lg: sizeModifierFn(theme, { fontSize: 16, radius: 24 }),
+    xs: sizeModifierFn(theme, sizeDict.xs),
+    sm: sizeModifierFn(theme, sizeDict.sm),
+    md: sizeModifierFn(theme, sizeDict.md),
+    lg: sizeModifierFn(theme, sizeDict.lg),
```
bstream: Currently these classes differ from those referenced on rows 93-96 (these are generated whereas the other ones are global)

PR #300, opened 2017-10-23
Feature/update alert

gish:


gish:
FYI @jesperwiner @robineng

PR #310, opened 2017-11-08
Feature/fund page changes

gish:
FYI @jesperwiner @robineng

### DIFF
--- a/src/components/tooltip/tooltip.jsx
+++ b/src/components/tooltip/tooltip.jsx
@@ -178,7 +178,7 @@ class Tooltip extends React.Component {
 }
 
 Tooltip.defaultProps = {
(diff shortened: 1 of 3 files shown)


gish:
Can I get a me-erge?

src/components/tooltip/tooltip.jsx
```diff
@@ -178,7 +178,7 @@ class Tooltip extends React.Component {
 }
 
 Tooltip.defaultProps = {
-  children: <Questionmark fill="#00A9EC" stroke="#00A9EC" width={16} height={16} />,
```
bstream: This will visually break onboarding process

bstream: After discussion it seems like this change works there too! Will get some input from design regarding hover state feedback and get back to you!

PR #311, opened 2017-11-09
feat: Add Shareville logo

JCB-K:
Add 2 version of the Shareville logo: with and without logotype

### DIFF
--- a/src/components/shareville-logo/index.jsx
+++ b/src/components/shareville-logo/index.jsx
@@ -1,0 +1,171 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+
+function SharevilleLogo({ onlyIcon, height }) {
+  const width = onlyIcon ? height : height * 4.89;
+
+  if (onlyIcon) {
+    return (
+      <svg viewBox="0 0 166.54 203.5" width={width} height={height}>
+        <defs>
+          <style
+            dangerouslySetInnerHTML={{
+              __html: '.svLogoGradientClass-1{fill:url(#linear-gradient);}.svLogoGradientClass-2{fill:url(#linear-gradient-2);}.svLogoGradientClass-3{fill:url(#linear-gradient-3);}',
+            }}
+          />
+          <linearGradient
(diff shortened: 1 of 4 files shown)


iamstarkov:
I would suggest to make it separate `SharevilleLogo` component

iamstarkov:
i mean without changin existing `Logo`

bstream:
I agree with @iamstarkov . The line 
```
All props except for `height` and `onlyIcon` are ignored.
```
in particular indicates that it should be handled differently and therefore it makes sense to have it in a separate component.

JCB-K:
@bstream @iamstarkov done

src/components/shareville-logo/index.jsx
```diff
@@ -0,0 +1,171 @@
...
+function SharevilleLogo({ onlyIcon, height }) {
+  const width = onlyIcon ? height : height * 4.89;
+
+  if (onlyIcon) {
+    return (
+      <svg viewBox="0 0 166.54 203.5" width={width} height={height}>
+        <defs>
+          <style
+            dangerouslySetInnerHTML={{
+              __html: '.svLogoGradientClass-1{fill:url(#linear-gradient);}.svLogoGradientClass-2{fill:url(#linear-gradient-2);}.svLogoGradientClass-3{fill:url(#linear-gradient-3);}',
```
codler: Why `svLogoGradientClass` as name? Is it autogenerated? Would be better if it contain shareville like the class below

JCB-K: @codler Fixed this now in a new release. (It's generated and never manually used by devs so not a big deal)

codler: no but next time someone generate svg it might get same name and will then conflict.

JCB-K: @codler ah I see what you mean. I already customised it by prefixing it with `svLogo`. (by default it was `gradientclass-1`, which is very generic). But good cal!

PR #313, opened 2017-11-14
feat(input): Add textarea component

JCB-K:


bstream:
Lets discuss this tmr.

src/components/input/input-default-styles.jsx
```diff
@@ -168,6 +168,10 @@ export default theme => {
         '&:focus': {
           outline: 'none',
         },
+        '&--textarea': {
+          maxHeight: 'none',
+          resize: 'vertical',
```
bstream: `resize` is not supported in IE - https://developer.mozilla.org/en-US/docs/Web/CSS/resize

src/components/input/input-default.jsx
```diff
@@ -99,37 +99,41 @@ class InputDefault extends React.PureComponent {
...
+      placeholder,
+      className: classes,
+      onFocus: this.onFocus,
+      onBlur: this.onBlur,
+      onChange: this.onChange,
+      value: this.state.value,
+    };
+
+    if (this.props.type === 'textarea') {
+      return <textarea {...props} style={{ height: this.props.lineCount * 21 }} />;
```
bstream: This style can be put in the class like so:
```
...: {
height: ({ lineCount }) => lineCount * 21,
}
```

Also, we might want to consider using something else than pixels so that fontSize can vary.

JCB-K: @bstream nice. fixed on both counts

PR #314, opened 2017-11-14
fix(Input): Reset input padding applied by some browsers

stigdanielek:
Set padding to 0 to prevent different input types having different browser-dependent default
paddings.

Before:
image

After:
image


stigdanielek:
Closed, because it did not work properly in Firefox. It created the opposite problem there. Back to the drawing board. :-)

PR #316, opened 2017-11-14
fix(Input): Reset input padding and set height

stigdanielek:
Set padding to 0 and define a height in order to make inputs have the same
height across browsers. Here are before-and-after screenshots from Chrome, Firefox and Internet Explorer:

image

The red lines are absolute-positioned behind the inputs to illustrate the difference between before (left column) and after (right column).

### DIFF
--- a/src/components/input/input-default-styles.jsx
+++ b/src/components/input/input-default-styles.jsx
@@ -160,10 +160,11 @@ export default theme => {
         width: '100%',
         transition: transitions.create(['border-color', 'transform']),
         fontSize: 14,
-        maxHeight: 27,
+        height: '1.5em',
         fontFamily: 'inherit',
         fontWeight: 'inherit',
         minWidth: 0,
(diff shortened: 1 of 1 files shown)


src/components/input/input-default-styles.jsx
```diff
@@ -161,9 +161,11 @@ export default theme => {
         transition: transitions.create(['border-color', 'transform']),
         fontSize: 14,
         maxHeight: 27,
+        height: 21,
```
bstream: Before we allowed inputs to resize depending on fontSize (Although up to a maximum of 27px for some reason). Lets check if this is still a valid use case!

bstream: Maybe we can change this to `em` and remove `maxHeight`?

stigdanielek: Good idea. I will change to that.

PR #318, opened 2017-11-15
ci(Travis): Dont build branches except master, only build pull requests

JCB-K:


JCB-K:
@iamstarkov as you can see Travis is still building the PR. But no unnecessary duplicate builds like before.

.travis.yml
```diff
@@ -6,8 +6,8 @@ node_js:
 
 # blacklist version commits
```
bstream: Change or remove this maybe? 🤔

bstream:
Maybe it is also time to drop Node 4 builds in travis. wdyt @iamstarkov @JCB-K ?

.travis.yml
JCB-K: @bstream whoops, fixed

PR #325, opened 2017-11-27
Fix/react warnings

bstream:
Fix react warnings in the documentation


codler:
The last test in badge still have `instrumentLvl`

PR #331, opened 2017-12-07
feat(icon): added exclamation point in circle

esundqvist:


bstream:
What do you think about reusing the ExclamationPoint icon and adding a border for the circle in the style clause?

Something like: 

```
{
    border: ({ fill }) => `1px solid ${fill}`, /* 1px could be calculated using width and height most probably */
    borderRadius: '50%',
    boxSizing: 'border-box',
}
```

PR #334, opened 2017-12-13
fix: center tooltip icon

johankitti:


src/components/tooltip/tooltip-styles.jsx
```diff
@@ -18,7 +18,8 @@ export default theme => {
 
     container: {
       cursor: 'pointer',
-      display: 'inline-block',
+      display: 'flex',
+      alignItems: 'center',
```
bstream: `alignItems` could probably be removed here, can you verify that?

PR #342, opened 2018-02-01
feat(Table): Add mobile list for tables

alinepickler:
New props: flexBasisMobile, flexBasisScreen, alignMobile, alignScreen, tableDataTitle, tableDataOrder

### DIFF
--- a/src/components/table/table.jsx
+++ b/src/components/table/table.jsx
@@ -16,11 +16,10 @@ function Table({
   ...rest
 }) {
   const usedClassName = classNames(classes.table, size, className);
-  const tableStyle = Object.assign({}, { minWidth }, style);
 
   return (
     <div className={classes.root}>
-      <table {...rest} style={tableStyle} className={usedClassName}>
+      <table {...rest} style={style} className={usedClassName}>
         {children}
       </table>
     </div>
@@ -33,7 +32,8 @@ Table.propTypes = {
   className: PropTypes.string,
   children: PropTypes.node,
   size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg']),
-  minWidth: PropTypes.number,
+  /** Use `'auto'` for mobile list mode */
+  minWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   style: PropTypes.object,
 };
 
--- a/src/components/td/td-styles.jsx
+++ b/src/components/td/td-styles.jsx
@@ -1,7 +1,7 @@
 import styleUtils from '../table/style-utilities';
 
 export default theme => {
-  const { palette, typography, mixins } = theme;
+  const { palette, typography, mixins, transitions } = theme;
 
   return {
     td: {
@@ -11,21 +11,91 @@ export default theme => {
       ...styleUtils.modifiers(palette),
       ...styleUtils.highlights(palette),
       ...styleUtils.borders(palette),
-      ...styleUtils.align(),
-      ...styleUtils.ellipsis(),
+
       fontFamily: typography.primary.fontFamily,
       fontWeight: 'inherit',
 
-      fontSize: 12,
+      padding: 0,
+      fontSize: 14,
       minWidth: 20,
-      paddingTop: 8,
-      paddingBottom: 8,
+      maxHeight: 50,
+      display: 'block',
+      transition: transitions.create(['max-height', 'border-width']),
 
       [mixins.media('md')]: {
-        fontSize: 14,
+        display: 'inline-block',
         minWidth: 40,
-        paddingTop: 10,
-        paddingBottom: 10,
+        border: 0,
+        '&::before': {
+          content: 'none',
+        },
+      },
+
+      '&.width': {
+        width: props => `${props.width}${typeof props.width === 'number' ? '%' : ''}`,
+      },
+
+      '&$collapsed': {
+        maxHeight: 0,
+        borderWidth: 0,
+
+        [mixins.media('md')]: {
+          maxHeight: 50,
+          borderWidth: 1,
+        },
+      },
+    },
+    collapsed: {},
+    ellipsis: {
+      whiteSpace: 'nowrap',
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+    },
+    hidden: {
+      display: props => (props.hiddenOnMobile ? 'none' : 'block'),
+      [mixins.media('md')]: {
+        display: props => (props.hiddenOnDesktop ? 'none' : 'inline-block'),
+      },
+    },
+    flexBasis: {
+      flexBasis: props => `${props.flexBasisMobile}${typeof props.flexBasisMobile === 'number' ? '%' : ''}`,
+
+      [mixins.media('md')]: {
+        flexBasis: props => `${props.flexBasisDesktop}${typeof props.flexBasisDesktop === 'number' ? '%' : ''}`,
+      },
+    },
+    child: {
+      display: 'block',
+      margin: [8, 6],
+      height: 'calc(100% - 16px)',
+
+      '&::before': {
+        display: 'block',
+        content: 'attr(data-title)',
+        fontSize: '0.8em',
+      },
+
+      [mixins.media('md')]: {
+        padding: [8, 6],
+        margin: 0,
+        height: '100%',
+        '&::before': {
+          content: 'none',
+        },
+      },
+    },
+    align: {
+      textAlign: props => (props.alignMobile ? props.alignMobile : props.align),
+
+      [mixins.media('md')]: {
+        textAlign: props => props.align,
+      },
+    },
+    flexOrder: {
+      order: props => props.flexOrder,
+
+      [mixins.media('md')]: {
+        order: () => 1,
       },
     },
   };
--- a/src/components/td/td.jsx
+++ b/src/components/td/td.jsx
@@ -12,6 +12,7 @@ function Td({
   size,
   width,
   mono,
+  collapsed,
   modifier,
   highlight,
   border,
@@ -19,8 +20,15 @@ function Td({
   borderRight,
   borderBottom,
   borderLeft,
+  alignMobile,
   align,
   ellipsis,
+  title,
+  flexOrder,
+  flexBasisMobile,
+  flexBasisDesktop,
+  hiddenOnDesktop,
+  hiddenOnMobile,
   theme, // eslint-disable-line react/prop-types
   sheet, // eslint-disable-line react/prop-types
   ...rest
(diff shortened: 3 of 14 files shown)


bstream:
👏 A bit swamped atm, will check this asap!

src/components/table/table.jsx
```diff
@@ -38,7 +37,7 @@ Table.propTypes = {
 };
```
FilipStenbeck: PropType for minWidth need to be PropTypes.oneOfType([PropTypes.number, PropTypes.string]),

alinepickler: Sure! I fixed it

src/components/td/td.jsx
```diff
@@ -77,8 +91,16 @@ Td.propTypes = {
   borderBottom: PropTypes.bool,
   borderLeft: PropTypes.bool,
   align: PropTypes.oneOf(['left', 'right', 'center']),
+  alignScreen: PropTypes.oneOf(['left', 'right', 'center']),
+  alignMobile: PropTypes.oneOf(['left', 'right', 'center']),
   /** By default a column will add ellipsis if the width is overflown. **Note:** this will only work if the child is a String. */
   ellipsis: PropTypes.bool,
+  tableDataTitle: PropTypes.string,
+  tableDataOrder: PropTypes.number,
```
bstream: What do you think about renaming `tableDataOrder` to `order` (Or possibly `flexOrder`?) and `tableDataTitle` to `title`?

src/components/td/td.jsx
```diff
@@ -77,8 +91,16 @@ Td.propTypes = {
   borderBottom: PropTypes.bool,
   borderLeft: PropTypes.bool,
   align: PropTypes.oneOf(['left', 'right', 'center']),
+  alignScreen: PropTypes.oneOf(['left', 'right', 'center']),
```
bstream: I am not sure about this `Screen` suffix naming. Does it make sense to use default values for mobile, eg. `align` for mobile and maybe `alignDesktop` for bigger screens?

(This applies to other props named `*Mobile` and `*Screen` as well)

src/components/th/th-styles.jsx
```diff
@@ -16,13 +16,22 @@ export default theme => {
...
 
-      minWidth: 20,
-      paddingBottom: 4,
-
       [mixins.media('md')]: {
         minWidth: 40,
         paddingBottom: 10,
       },
+
+      '&.flexBasis': {
```
bstream: What do you think about converting these to standalone classes instead? Then they will get generated names from jss.

bstream:
While they are not really related to this PR it would also be nice if we could clean up some `!important` statements :)

src/components/table/table.jsx
```diff
@@ -33,12 +32,12 @@ Table.propTypes = {
...
   children: PropTypes.node,
   size: PropTypes.oneOf(['xs', 'sm', 'md', 'lg']),
-  minWidth: PropTypes.number,
+  minWidth: PropTypes.oneOfType([PropTypes.number, PropTypes.string]),
   style: PropTypes.object,
 };
 
 Table.defaultProps = {
-  minWidth: 700,
+  minWidth: 'auto',
```
mlenser: If no `minWidth` is provided we likely shouldn't add extra css tags unless `auto` is desired here. I assume it is added to prevent the `table-styles` code from breaking without a `minWidth`? If so, then we should only apply that style if `minWidth` is specified.

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
       minWidth: 20,
       paddingTop: 8,
       paddingBottom: 8,
+      display: 'block',
```
mlenser: A `<td>` should likely never be set as `block` as they are normally `table-cell`. Perhaps in mobile, but we'd likely not use a table then...

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
...
+        [mixins.media('md')]: {
+          order: () => 1,
+        },
+      },
+
+      '&.flexBasis': {
+        flexBasis: props =>
+          props.flexBasisMobile
+            ? `${props.flexBasisMobile}${typeof props.flexBasisMobile === 'number' ? '%' : ''}`
+            : `${props.flexBasisDesktop}${typeof props.flexBasisDesktop === 'number' ? '%' : ''}`,
```
mlenser: Why assume percentages here? `flex-basis` can accept many value types and reading the prop `flexBasisMobile` doesn't indicate percentage. Why not pass in the percentage via the prop?

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
...
+
+        [mixins.media('md')]: {
+          flexBasis: props => `${props.flexBasisDesktop}${typeof props.flexBasisDesktop === 'number' ? '%' : ''}`,
+        },
+      },
+
+      '&.width': {
+        width: props => `${props.width}${typeof props.width === 'number' ? '%' : ''}`,
```
mlenser: Same question as `flex-basis` above

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
...
+      '&.align': {
+        textAlign: props => (props.alignMobile ? props.alignMobile : props.alignDesktop || props.align),
+
+        [mixins.media('md')]: {
+          textAlign: props => props.alignDesktop || props.align,
+        },
+      },
+
+      '&.hidden': {
+        display: props => (props.hiddenOnMobile ? 'none' : 'block'),
```
mlenser: Another dev mentioned this can be deconstructed:
```
 display: ({hiddenOnMobile}) => (hiddenOnMobile ? 'none' : 'block'),
```

as above, `block` should be `table-cell`, though I'd much prefer finding a way to only set the `display` prop if `hiddenOnMobile` is set, though I can't find examples of such a case.

src/components/tr/tr-styles.jsx
```diff
@@ -9,17 +9,17 @@ export default theme => {
...
       ...styleUtils.flexRow(),
       ...styleUtils.sizes(),
-      alignItems: 'center',
       borderLeft: '1px solid transparent',
       borderRight: '1px solid transparent',
+      flexWrap: 'wrap',
 
       '&.primary': {
-        background: `${color.grayLight} !important`,
+        background: () => `${color.grayLight}`,
```
mlenser: Why add the fat arrow? Is it to get it more specific?

Ideally this should be `backgroundColor: color.grayLight`

src/components/tr/tr-styles.jsx
```diff
@@ -31,22 +31,22 @@ export default theme => {
...
-      [mixins.media('md')]: {
-        '&.sticky': {
-          position: 'fixed',
-        },
-
-        '& + .tr--clone': {
-          display: 'flex',
+        [mixins.media('md')]: {
+          '&.sticky': {
+            position: () => 'fixed',
```
robineng: unnecessary function

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
...
+
+      '&.width': {
+        width: props => `${props.width}${typeof props.width === 'number' ? '%' : ''}`,
+      },
+
+      '&.align': {
+        textAlign: props => (props.alignMobile ? props.alignMobile : props.alignDesktop || props.align),
+
+        [mixins.media('md')]: {
+          textAlign: props => props.alignDesktop || props.align,
```
robineng: This will break in IE11 due to a bug with dynamic props and nested jss rules mantioned here: https://github.com/cssinjs/jss/issues/623

src/components/td/td-styles.jsx
```diff
@@ -20,12 +20,61 @@ export default theme => {
...
+        },
+      },
+
+      '&.hidden': {
+        display: props => (props.hiddenOnMobile ? 'none' : 'block'),
+
+        [mixins.media('md')]: {
+          display: props => (props.hiddenOnDesktop ? 'none' : 'inline-block'),
+        },
```
mlenser: I think this could be:
```
hiddenOnDesktop: {
  [mixins.media('md')]: {
    display: 'none',
  },
},
hiddenOnMobile: {
  [mixins.media('sm')]: {
    display: 'none',
  },
},
```

src/components/td/td.jsx
```diff
@@ -30,23 +38,26 @@ function Td({
     size,
     {
       hasWidth: width,
+      width,
+      flexBasis: flexBasisMobile || flexBasisDesktop,
       [modifier]: modifier,
       [`highlight-${highlight}`]: highlight,
-      [`align-${align}`]: align,
+      align: alignMobile || alignDesktop || align,
+      hidden: hiddenOnMobile || hiddenOnDesktop,
```
mlenser: This could be:
```
[classes.hiddenOnDesktop]: hiddenOnDesktop,
[classes.hiddenOnMobile]: hiddenOnMobile,
```

src/components/tr/tr-styles.jsx
alinepickler: Needed due to dynamic stylesheet overrides static ones

src/components/table/table.jsx
alinepickler: removed

src/components/td/td-styles.jsx
alinepickler: Can you explain more ? If its a number its percent, if its string its pixels, just like all the others like this one all over the UI kit

src/components/td/td-styles.jsx
alinepickler: See above

src/components/tr/tr-styles.jsx
alinepickler: Fixed

src/components/td/td-styles.jsx
alinepickler: No, If I use :
  [mixins.media('sm')]: {
    display: 'none',

it will hide for all SM and above (MD, LG) so  including PC screens

src/components/td/td.jsx
alinepickler: see above

src/components/td/td-styles.jsx
alinepickler: We'll fix it, we see the problem. Toggle on classname instead of nested dynamic props

src/components/td/td-styles.jsx
mlenser: Right, we need a mixin to target mobile specifically if we were to do this implementation. Such a solution seems a better path. In robosave we use
```
onMobileOnly: `@media (max-width: ${theme.breakpoints.sm - 1}px)`,
onDesktop: `@media (min-width: ${theme.breakpoints.sm}px)`,
```
@bstream what would you think of doing the same in `ui-kit`?

mlenser: Captured as part of #346

src/components/td/td-styles.jsx
mlenser: Captured as part of #347

src/components/td/td-styles.jsx
```diff
@@ -12,20 +12,93 @@ export default theme => {
...
       fontFamily: typography.primary.fontFamily,
       fontWeight: 'inherit',
 
-      fontSize: 12,
+      padding: 0,
+      fontSize: 14,
       minWidth: 20,
-      paddingTop: 8,
-      paddingBottom: 8,
+      maxHeight: 200,
```
bstream: Should this be something more dynamic, maybe `100%` or similar?

src/components/td/td-styles.jsx
```diff
@@ -12,20 +12,93 @@ export default theme => {
...
+          content: 'none',
+        },
+      },
+
+      '&.collapsed': {
+        maxHeight: 0,
+        borderWidth: 0,
+
+        [mixins.media('md')]: {
+          maxHeight: 200,
```
bstream: Where does this `200` come from?

src/components/th/th-styles.jsx
```diff
@@ -15,14 +15,23 @@ export default theme => {
...
+
+      '&.width': {
+        width: props => `${props.width}${typeof props.width === 'number' ? '%' : ''}`,
+      },
+    },
+    flexBasis: {
+      flexBasis: props => `${props.flexBasisDesktop}${typeof props.flexBasisDesktop === 'number' ? '%' : ''}`,
+    },
+    align: {
+      textAlign: props => props.alignDesktop || props.align,
```
bstream: Why are we defaulting this to desktop?

bstream:
Note to self: Clean up commit messages before merging :)

Issue #229, opened 2017-05-23
Add IconButton


bstream:
![image](https://cloud.githubusercontent.com/assets/3505504/26353774/3dbbeca2-3fc2-11e7-8472-f1e459df48fb.png)

mlenser:
Closed as part of #343

PR #343, opened 2018-02-01
feat(Button): add round icon buttons

mlenser:
**Purpose:**
This PR addresses a gap between design and the ui-kit. Design wants round buttons for icons as seen https://next.test.nordnet.se/overview. This PR should allow all projects to use a standard setup for icon buttons and not create their own styling.

**Uncertainties:**

- Duplication of styles. There were many cases of duplication for the `boxShadow` so I put the svg `transform` next to all of those. It seems a bit messy in the current setup, but I've just followed that setup.
- `1.5em` on the icons in the button. We need to use `em` here to inherit from the button `fontSize`, but I'm unsure of the desired sizes of the icons. Perhaps icons should even be `em` by default with the option of setting a size. Feedback is welcome here.

Please let me know if there is any feedback!

**Screenshots:**

icon button primary
icon button secondary
icon button link
icon button sizes


bstream:
👏Thanks for this! Will check back tomorrow. Also, when this is merged #229 can be closed 👍

src/components/button/button-styles.jsx
```diff
@@ -23,6 +23,9 @@ const buttonModifierFn = (variant, color, colorFocus, colorDisabled, textColorDi
 
       '&:focus, &:active': {
         boxShadow: `0 0 0 2px ${colorFocus}`,
+        '& svg': {
```
bstream: What do you think about "protecting" these `transform`s so that we only use them in `iconButton`? These will affect all svgs inside the button atm (say you put an svg inside a normal button without `iconButton` being set)

mlenser: I find the SVG zooming behavior to be weird in the first place. We already have the outer border change so the active state is apparent without the scaling of the svg. Which designer would be the person to talk about the desired behavior of these buttons?

If that person says that we should keep it then we could implement them so that they only impact `iconButton` buttons.

bstream: Fully agree, lets find someone and sync!

mlenser:
Had some conversations on Friday and we've removed the svg scale - great!

I'll also try to tackle icons and text in the same buttons. I have a good setup now, but need to consider how much the heights of each button type (text, icon, icon+text) should align. I've posed the question to @There5a and we'll follow up on Monday.

Here is the current situation which still needs some massaging:

icon buttons ideas

mlenser:
The setup is now updated to handle text, icon+text, and icon. All heights are set as desired by design:

buttons sized

Icon's default color and sizes have been changed - I believe they are better defaults, but if needed could handle it in another way.

src/components/icon/icon.jsx
```diff
@@ -23,13 +23,13 @@ Icon.propTypes = {
...
-  fill: '#222',
-  stroke: '#222',
+  fill: 'currentColor',
+  stroke: 'currentColor',
   'aria-hidden': 'true',
   focusable: 'false',
   strokeWidth: 1,
-  height: 16,
-  width: 16,
+  height: '1em',
```
bstream: Unfortunately this will break the `Ellipse` icon. However I think that it is better to default it to `1em` like you did here so to fix the error we can add `height: 16, width: 16` to `Ellipse` defaultProps instead!

src/components/button/button.md
```diff
@@ -53,6 +53,23 @@ Secondary, disabled:
       <Button variant="secondary" disabled modifier="warning" style={ style }>Click Me</Button>
       <Button variant="secondary" disabled modifier="success" style={ style }>Click Me</Button>
     </div>
+    
+Secondary, icon:
```
bstream: No `primary` icons?

src/components/button/button-styles.jsx
```diff
@@ -106,27 +106,88 @@ export default theme => {
...
+        width: 18,
+      },
+    },
+
+    '.icon': {
+      borderRadius: '50%',
+    },
+
+    innerWrapper: {
+      alignItems: 'center',
```
bstream: What do you think about adding `justifyContent: center` as well?

src/components/button/button-styles.jsx
```diff
@@ -106,27 +106,88 @@ export default theme => {
...
+      '&.icon': {
+        padding: 10,
+      },
+      '& svg': {
+        height: 18,
+        width: 18,
+      },
+    },
+
+    '.icon': {
```
bstream: You can use JSS for these *(icon, iconText)* also. To refer to them inside other elements (such as `'&.icon'`) you simply use `'&$icon'` instead :)

src/components/button/button.md
mlenser: I removed them as @There5a said we shouldn't use them. They're still available, just not in the documentation to avoid confusion

src/components/button/button-styles.jsx
mlenser: It won't do anything as the inner is just a wrapper around text+icon. That element is already centered by the padding. Is there a reason to add it?

src/components/icon/icon.jsx
mlenser: Fixed on Ellipse and no console errors when viewing it locally so this should be ok now.

src/components/button/button-styles.jsx
mlenser: Oooh, I didn't know about `'&$icon'`. I originally had them under `classes` and was searching around for how to make it work, but settled on following the current pattern. I'll refactor to that!

mlenser: The only issue is that `iconText` has no default styles, only styles for each size. I could implement it with an empty prop:
```
iconText: {},
```
Seems a bit hacky, but it'd work.

src/components/button/button-styles.jsx
bstream: Ah, yeah that's true. No reason to add it then :)

Issue #336, opened 2017-12-18
New yellow colour


There5a:
Please update yellow colour to this new one: #FFD75A (old: #E5AA4C)
Won't work as text colour, so change alert text to black (new design for alerts will come soon!).

I'm not sure if the yellowLight is used. If so, keep it as is for now. If not, kill it :)

bstream:
Closed with #344

PR #344, opened 2018-02-02
fix(Style): update yellow color

mlenser:
Addresses the main part of #336. I've left `yellowLight` as I'm unable to determine if it's used and removing it may be considered a breaking change.

### DIFF
--- a/src/styles/palette.js
+++ b/src/styles/palette.js
@@ -4,7 +4,7 @@ const variant = {
   primary: color.blue,
   info: color.blueDark,
   success: color.green,
-  warning: color.yellow,
+  warning: color.black,
   danger: color.red,
 };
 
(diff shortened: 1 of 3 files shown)


src/styles/palette.js
```diff
@@ -4,7 +4,7 @@ const variant = {
   primary: color.blue,
   info: color.blueDark,
   success: color.green,
-  warning: color.yellow,
+  warning: color.black,
```
bstream: I think that this one should be kept as yellow but in the `Alert` component the text should be changed to black instead of yellow, which it is today. This can be done like this: 
```

mlenser: Ah, I didn't notice buttons broke. I checked them on my local, but likely before I replaced the `palette.warning`.

Should be fixed with my latest push:
black text warning alert

warning buttons

PR #345, opened 2018-02-07
feat(fund-explanation): added graphics for fund explanation

esundqvist:


bstream:
After discussing we decided that this is not *yet* a worthy candidate for the ui-kit as it is just used in one place at the moment. Will reevaluate this if that changes!

PR #351, opened 2018-02-08
fix(Buttons): only size svgs provided via the `icon` prop

mlenser:
Prevent it from modifying normal icons in the button:

normal icon in button


mlenser:
Thanks!

PR #352, opened 2018-02-12
fix(Tooltip): Prevent tooltip from going off screen on mobile

### LINKED ISSUE #339: Prevent tooltip from going off the screen on mobile
Similar to #327 the tooltip goes off the page a bit, but on mobile. This only occurs on a tooltip places slightly right of center. Another tooltip placed dead center does not have this issue.

It's pre-rendered so it creates a bit of a horizontal scroll bar even when not hovered (I've scrolled over, the darker gray is normally not so visible).

The tooltip is using a fixed width of 300 and the mobile size is 320.

screen shot 2018-01-16 at 15 00 26

bstream:
Solves the case when a tooltip should fit (although barely) but gets rendered off screen. This happens since:
1. The default space from the edge (10) is to big
2. The meta viewport tag is invalid, which makes `window.innerWidth` to behave badly.

Fixes #339

### DIFF
--- a/documentation/template.html
+++ b/documentation/template.html
@@ -4,7 +4,7 @@
 		<meta charset="utf-8">
 		<title><%=htmlWebpackPlugin.options.title%></title>
     <link href="https://fonts.googleapis.com/css?family=Bitter:400,700|Open+Sans:300,400,600,700,800" rel="stylesheet">
-		<meta name="viewport" content="width=device-width, initial-scale=1.0">
+		<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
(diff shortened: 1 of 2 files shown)


mlenser:
I tested it a bit more:

screen shot 2018-02-13 at 10 04 02

This scenario occurs using this branch and normal ui-kit. The tooltip increases the right side of the page's gutter to accomodate it. It seems it's no longer adding a horizontal scroll bar (on either master or this branch) so it's no longer as severe of an issue, but there seem to still be some issues.

Either way this code looks like an improvement and I'd say the priority for this behavior is very low.

PR #354, opened 2018-02-15
Button fixes

bstream:


src/components/button/button.md
```diff
@@ -4,12 +4,15 @@ Primary button:
...
-    <div>
+    <div style={{display: 'flex'}}>
       <Button variant="primary" style= { style }>Click Me</Button>
       <Button variant="primary" modifier="action" style= { style }>Click Me</Button>
       <Button variant="primary" modifier="danger" style= { style }>Click Me</Button>
       <Button variant="primary" modifier="warning" style= { style }>Click Me</Button>
       <Button variant="primary" modifier="success" style= { style }>Click Me</Button>
+      <div style={{width: 300}}>
+        <Button block variant="primary" modifier="action" style= { style }>Click Me</Button>
+      </div>
```
mlenser: Is this something we want to allow? We encountered it and I thought about it, but it seems like something that we should discourage or entirely disallow. Material design for example doesn't allow customization of width.

bstream: That is a good question, I am not sure. However, this is not any new functionality I am adding, I am just fixing it since it was broken. If we want to remove that it is a breaking change

I am not sure about the example of material design; they seem to allow block buttons as well (called fullWidth there) https://github.com/mui-org/material-ui/blob/v1-beta/src/Button/Button.js#L145

mlenser: Interesting, they don't have `fullWidth` in their documentation, only in the code.

Something of note is that `material-ui` is not made by Google. Material-ui attempts to implement the design of material design into react and I really like the project and use it myself (v1), but it isn't always the same in small details.

This is a good fix if we don't want to make a breaking change. In the long run it would be good to get alignment from design on consistency of button sizing.

codler:
👍

PR #356, opened 2018-02-19
fix(table): th has some of the same fields as td

mlenser:
`td` already supports these fields - added support for them on `th`


mlenser:
Table is possibly being rewritten so this is on hold.

mlenser:
#358 will handle this

PR #357, opened 2018-02-19
fix(tooltip): color should inherit from parent as all icons do

mlenser:
With the change that was made to icons they are now set to inherit from their color from their parent. This syncs that same behavior.

I'm following up with design for if words/svgs should have any color when they have a tooltip and will handle that separately if it changes.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ tmp
 .tmp
 coverage
 .nyc_output
+.idea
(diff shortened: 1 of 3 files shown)


mlenser:
Verified with design that the icon should not be colored and that text isn't used. We should potentially remove the ability to pass text to tooltip as well, but this will meet the requirements.

PR #359, opened 2018-02-26
fix(LabeledValue): Fix incorrect padding calculation

gish:
The current implementation resulted in some paddings being NaNpx


bstream:
What do you think about doing like this instead? #360

gish:
Yes, way less convoluted.

bstream:
Cool, then I'll close this PR and check the other one with design on monday ;)

PR #362, opened 2018-03-02
fix: Add SolidArrowUp & SolidArrowDown icons

geordidearns:


test/components/instrument-badge/instrument-badge.test.js
```diff
@@ -33,14 +33,9 @@ describe('<InstrumentBadge />', () => {
...
   });
 
-  it('should render close icon when NOT marked as qualified', () => {
+  it('should render ExclamationPoint icon when NOT marked as qualified', () => {
     wrapper = shallow(<InstrumentBadge classes={classes} theme={theme} />);
     expect(wrapper.find(Checkmark).length).to.equal(0);
     expect(wrapper.find(ExclamationPoint).length).to.equal(1);
   });
-
-  it('should render instrument level and subtext', () => {
```
bstream: Add this test case back please :)

src/components/icon/icons/solidArrowDown.jsx
```diff
@@ -0,0 +1,25 @@
+import PropTypes from 'prop-types';
+import React from 'react';
+import SolidArrowUp from './solidArrowUp';
+
+export default function SolidArrowDown({
+  style: styleProp,
+  ...rest // eslint-disable-line comma-dangle
```
mlenser: Is there a reason to disable the comma dangle for the spread operator? I tried removing it and it doesn't seem to cause any issues, though perhaps I've overlooked it.

geordidearns: @mlenser sorry what do you mean?

mlenser: I'm asking about `// eslint-disable-line comma-dangle`. I would expect this to not have the eslint-disable like such:

```
export default function SolidArrowDown({
  style: styleProp,
  ...rest,
})
```
It seemed to work when I made that change, but maybe you saw something else.

geordidearns: Yeah cool, will update

mlenser: Nice!

mlenser:
Looks good to me overall.

As an optimization it'd probably be better to have one generic arrow and then extend that generic arrow 4 times for up, down, left, and right. And in the generic arrow handle the translate (possibly in the SVG instead of the css). That would avoid extra inline styles.
Extra inline styles don't matter so much at small scale, but at the medium scale it starts to matter more as they add up. For example if we have a series of data with 200 entries and each entry has an arrow then we now have 200 more inline styles, thus slowing the performance of our app.

mlenser:
Captured as https://github.com/nordnet/nordnet-ui-kit/issues/363

geordidearns:
@mlenser - It's also across a lot of other components, It's a fight between doing what's right now (Which I agree on the above) and getting something out to the user today which is due. Perhaps when someone takes a look at this one they can look at going through all the components too.

mlenser:
@geordidearns agreed, wasn't expecting it to be fixed as part of this. Nice work. :)

PR #364, opened 2018-03-02
fix(table): Add limited width example and never allow the table cells to touch

mlenser:
Added an example and padding left/right. Without left/right padding the cells can touch eachother if there is too much content. There doesn't seem to be any issue with the left/right padding on the other examples.

### DIFF
--- a/src/components/table-cell/table-cell-styles.jsx
+++ b/src/components/table-cell/table-cell-styles.jsx
@@ -14,7 +14,7 @@ export default theme => {
       fontFamily: typography.primary.fontFamily,
       fontWeight: 'inherit',
 
-      padding: 0,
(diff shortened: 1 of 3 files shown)


src/components/table-cell/table-cell-styles.jsx
```diff
@@ -63,7 +63,6 @@ export default theme => {
     },
     child: {
       display: 'block',
-      margin: [8, 0],
```
jesperwiner: Can't have padding on TD/TH due to flexbasis bug in IE11, see https://github.com/philipwalton/flexbugs#flexbug-7

I've tested and should be a margin: 8 on child and padding: 0 on the tablecell to get same effect

Issue #340, opened 2018-01-30
v2


bstream:
- [x] Remove deprecated APIs #368
- [x] Move some dependencies to `peerDependencies` (for example `react`) #369
- [x] react-jss #317, #370
- [x] Table refactoring #371

lets cut it here

bstream:
Take a closer look at the following components:

* Table
* Button
* Input

iamstarkov:
PR 372

PR #368, opened 2018-03-07
v2 / remove deprecations #340

iamstarkov:
refactor: remove built-in ThemeProvider from jss-theme-reactor

It was fun, but its over. Please use ThemeProvider from react-jss.

BREAKING CHANGE: ThemeProvider is removed along with jss-theme-reactor. See here
https://github.com/nordnet/nordnet-ui-kit/blob/v1.30.0/docs/migrations/v1.7.0.md

----

refactor: remove deprecated test utils, like createShallow, createMount and createRenderToString

BREAKING CHANGE: if you didnt follow 1.7 migration guides to use enzyme then its about the time /docs/migrations/v1.7.0.md@v1.30.0


iamstarkov:
more deprecations are coming

PR #377, opened 2018-03-20
feat: Add Feed icon

JCB-K:
Add icon for social feeds


src/components/icon/icons/index.js
```diff
@@ -1,84 +1,85 @@
...
+import SocialInstagram from "./socialInstagram";
+import SocialTwitter from "./socialTwitter";
+import SocialYouTube from "./socialYouTube";
+import SolidArrowDown from "./solidArrowDown";
+import SolidArrowUp from "./solidArrowUp";
+import Star from "./star";
+import Tag from "./tag";
+import TickingClock from "./tickingClock";
+import Trash from "./trash";
+import VerticalEllipsis from "./verticalEllipsis";
```
mlenser: Why double quotes? :(

JCB-K: @mlenser My answer could be "because I didn't pay attention" or "because Prettier config was broken" :). I made a new PR where I fix it, if we merge that one first and then this one we'll be good.

JCB-K:
The diff one this one is confusing until #378 lands on master

src/components/icon/icons/feed.jsx
```diff
@@ -0,0 +1,53 @@
...
+  ...rest // eslint-disable-line comma-dangle
+}) {
+  const style = {
+    ...styleProp,
+    ...Icon.defaultProps.style,
+  };
+  return (
+    <svg style={style} {...rest}>
+      <g stroke="none" strokeWidth={strokeWidth} fill={fill} fillRule="evenodd">
+        <g fill={fill}>
```
bstream: You should be able to remove this `<g>` since there is already a `fill` on its parent

src/components/icon/icons/feed.jsx
```diff
@@ -0,0 +1,53 @@
...
+            stroke={fill}
```
bstream: This should probably be `stroke` instead of fill?

JCB-K:
thanks @bstream, fixed

PR #381, opened 2018-04-04
docs: Add Theme Visualiser

### LINKED ISSUE #380: Add colors to the documentation
It would be nice to have the colors available listed in the documentation in some way so you don't have to find it through the code/github.

JCB-K:
Resolves #380 

This adds a component to the Styleguide page that shows all colors in the theme. Could easily be extended with fonts, animations, etc.


bstream:
@JCB-K Thanks! I broke out theme doc so it won't end up in `dist`. Also separated Components from Theme in the docs and made a few adjustments.

PR #385, opened 2018-04-26
fix: don't default node to button

bstream:


mlenser:
Doesn't seem to work. I tested it locally and the buttons are still `<button>` whereas I'd expect an `<a>`

bstream:
The old logic looked like this: `const Element = href ? 'a' : 'button';` so that is why we did like this. (i.e. you only get a `<a>` if you have an `href` (or in this case if you put `node="a"`))

PR #386, opened 2018-04-26
feat(popup-menu): adds popup menu

esundqvist:
adds popup menu previously used only in page-watchlist

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
     "classnames": "^2.2.3",
     "color": "^1.0.3",
     "lodash": "^4.17.4",
-    "react-transition-group": "^2.2.0"
+    "react-transition-group": "^2.3.1"
   },
   "peerDependencies": {
     "prop-types": ">=15",
@@ -124,6 +124,7 @@
     "react": "^15.5.4",
     "react-dom": "^15.5.4",
     "react-jss": "^8.3.3",
+    "react-router-dom": "^4.2.2",
     "react-styleguidist": "^5.5.10",
     "react-test-renderer": "^15.5.4",
     "rimraf": "^2.6.1",
--- a/src/components/popup-menu/popup-menu-item.jsx
+++ b/src/components/popup-menu/popup-menu-item.jsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import injectSheet from 'react-jss';
+import styles from './popup-menu.styles';
+
+function PopupMenuItem({ node, children, topBorder, linkTo, onClick, disabled, classes }) {
+  const Element = node || 'button';
+  return (
+    <li className={classes.item}>
+      {topBorder && <hr className={classes.hr} />}
+      <Element to={linkTo} onClick={onClick} className={classes.link} disabled={disabled}>
+        {children}
+      </Element>
+    </li>
+  );
+}
+
+PopupMenuItem.propTypes = {
+  classes: PropTypes.object.isRequired,
+  node: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  onClick: PropTypes.func,
+  linkTo: PropTypes.string,
+  topBorder: PropTypes.bool,
+  children: PropTypes.oneOfType([PropTypes.node, PropTypes.string, PropTypes.number]),
+  disabled: PropTypes.bool,
+};
+
+PopupMenuItem.defaultProps = {
+  onClick: () => {},
+  children: '',
+  linkTo: '#',
+  topBorder: false,
+  disabled: false,
+};
+
+export default injectSheet(styles)(PopupMenuItem);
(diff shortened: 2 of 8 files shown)


package.json
```diff
@@ -75,7 +75,8 @@
     "classnames": "^2.2.3",
     "color": "^1.0.3",
     "lodash": "^4.17.4",
-    "react-transition-group": "^2.2.0"
+    "react-router-dom": "^4.2.2",
```
bstream: Could we make this something that the user sends in instead? (Would like to keep out "specialized" dependencies such as this from the ui-kit and keep it general)

src/components/popup-menu/popup-menu-item.jsx
```diff
@@ -0,0 +1,53 @@
...
+  onClick: () => {},
+  button: false,
+  children: '',
+  linkTo: '#',
+  topBorder: false,
+  className: '',
+  disabled: false,
+};
+
+export default injectSheet(styles)(PopupMenuItem);
```
bstream: Related to the comment in package.json: Rather than sending in button we might want to give the user the possibility to send in a `node` instead. We recently introduced this concept in the `<Button />` component here: https://github.com/nordnet/nordnet-ui-kit/pull/384/files. What do you think?

src/components/popup-menu/popup-menu.jsx
```diff
@@ -0,0 +1,104 @@
...
+      >
+        <button className={classes.menuButton} onClick={this.toggleMenu}>
+          {toggleButton || <Icon.VerticalEllipsis height={28} width={28} stroke="currentColor" fill="currentColor" />}
+        </button>
+        <TransitionGroup>
+          {isOpen && (
+            <CSSTransition classNames={classes.menuSlideDown} timeout={{ exit, enter }}>
+              <div className={classes.menuPopup} style={{ width }}>
+                <div className={classes.menuItemContainer}>
+                  <div>
```
bstream: Is it necessary to have this extra div here?

src/components/popup-menu/popup-menu.jsx
```diff
@@ -0,0 +1,104 @@
...
+            </CSSTransition>
+          )}
+        </TransitionGroup>
+      </span>
+    );
+  }
+}
+
+PopupMenu.propTypes = {
+  width: PropTypes.number,
```
bstream: Would it make sense that this could be a string as well? (I am thinking about these cases: `auto` or `100%`)

src/components/popup-menu/popup-menu.styles.js
```diff
@@ -0,0 +1,111 @@
...
+    position: 'absolute',
+    display: 'inline-block',
+    top: 0,
+    right: 0,
+    background: palette.color.white,
+    boxShadow: '0 2px 4px 2px rgba(0, 0, 0, .1)',
+    listStyle: 'none',
+    padding: 15,
+    marginTop: 55,
+    zIndex: 10,
```
bstream: `zIndex: 1` should be enough since it is local

src/components/popup-menu/popup-menu.styles.js
```diff
@@ -0,0 +1,111 @@
+export default ({ palette, typography, mixins }) => ({
```
bstream: Has this been visually tested in more than one browser? What about IE11?

bstream: Does this support changing the font-size of the menu?

src/components/popup-menu/popup-menu.styles.js
```diff
@@ -0,0 +1,111 @@
...
+  menuSlideDown: {
+    '&-enter': {
+      opacity: 0.01,
+      transform: 'translate(0, -5px)',
+
+      '&-active': {
+        opacity: 1,
+        transform: 'translate(0, 0)',
+        transitionProperty: 'transform, opacity',
+        transition: 'opacity 300ms ease-in',
```
bstream: Create transitions using the built-in helper for consistency.
`transition: transitions.create(['opacity', 'transform']),`

package.json
esundqvist: I understand, but I don't see an easy solution for handling the states of each <PopupMenuItem /> without using <Link />. We would move a lot of logic to the component implementing <PopupMenu /> and bloat those instead.

src/components/popup-menu/popup-menu-item.jsx
esundqvist: Not sure I understand, accepted children props are PropTypes.oneOfType([PropTypes.node, PropTypes.string, PropTypes.number])

src/components/popup-menu/popup-menu.jsx
esundqvist: Nope :)

src/components/popup-menu/popup-menu.jsx
esundqvist: Absolutely correct.

src/components/popup-menu/popup-menu.styles.js
esundqvist: Tested in Chrome, Safari, Firefox, IE11

esundqvist: Nope.

PR #388, opened 2018-05-04
feat(popup-menu): allow maxHeight for popupMenu

esundqvist:


src/components/popup-menu/popup-menu.jsx
```diff
@@ -92,6 +93,7 @@ PopupMenu.propTypes = {
   enter: PropTypes.number,
   exit: PropTypes.number,
   toggleButton: PropTypes.node,
+  maxHeight: PropTypes.string,
```
bstream: What do you think about allowing numbers here as well?

PR #390, opened 2018-05-18
fix(button): fix disable for button with href

KitsuneKelso:
button will render as button element rather than anchor when supplied with both href and disabled


KitsuneKelso:
There was a bug where passing an `href` to the button but also passing `disabled` would not disable the actual button, since it was being rendered as an `<a>` rather than a `<button>` for all situations where it was given an `href`. This seemed like the simplest way to fix the problem.

PR #391, opened 2018-05-21
Fix too many dynamic stylesheets for tooltip

JacobBlomgren:
Fixes MIG-4733. Tested in modern browsers and IE, it shouldn't break anything.


bstream:
Thanks! 🎉

PR #395, opened 2018-05-31
fix too many dynamic stylesheets for table

JacobBlomgren:
Doesn't break the api, nor any functionality. I simply wrap the `Tr` and `Thead` components in a component that injects dynamic stylesheets only when needed (when `stickyOffset` is used, which it hardly is, but that's different problem). 

Tested in modern browsers and works as before. It works just as before in IE as well, the problem is that `sticky` is not [supported](https://caniuse.com/#feat=css-sticky) in IE so it has never worked there.

Perhaps the most important of all the stylesheet fixes. Will dramatically improve performance as we use tables so frequently.

### DIFF
--- a/package.json
+++ b/package.json
@@ -74,7 +74,7 @@
   "dependencies": {
     "classnames": "^2.2.3",
     "color": "^1.0.3",
-    "lodash": "^4.17.4",
+    "lodash": "^4.17.10",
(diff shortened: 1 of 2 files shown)


JacobBlomgren:
:flushed: :flushed: :flushed: :flushed:

PR #397, opened 2018-06-04
feat(Icon): Add icons MagnifierPlus and MagnifierMinus

anncos:


src/components/icon/icons/magnifierMinus.jsx
```diff
@@ -0,0 +1,41 @@
...
+  style: styleProp,
+  ...rest // eslint-disable-line comma-dangle
+}) {
+  const style = {
+    ...styleProp,
+    ...Icon.defaultProps.style,
+  };
+  return (
+    <svg style={style} {...rest}>
+      <g id="magnifierMinus" stroke="none" strokeWidth={strokeWidth} fill="none" fillRule="evenodd">
```
JCB-K: the ideas are not required I think?

anncos: Thanks! It seems like you are right

PR #398, opened 2018-06-04
Feat/add carousel

mjmonline:
Extracts Carousel from market-performance-widget, to be reused on Discovery page


mjmonline:
https://scm.prod.nordnet.se/projects/MIG/repos/market-performance-widget/browse

src/components/carousel/carousel-styles.jsx
```diff
@@ -0,0 +1,65 @@
...
+  },
+  button: {
+    cursor: 'pointer',
+    color: palette.color.grayDarker,
+    outline: 'none',
+    background: 'transparent',
+    width: 30,
+    height: 30,
+    border: 'none',
+    margin: '0 0 0 8px',
```
iamstarkov: `margin: [0, 0, 0, 8]`

src/components/carousel/carousel-styles.jsx
```diff
@@ -0,0 +1,65 @@
+export default ({ palette, mixins }) => ({
+  wrapper: {
+    display: 'flex',
+    minHeight: 100,
+  },
+  swipeContainer: {
+    overflow: 'hidden',
+    margin: 'auto',
+    width: '100%',
+    maskImage: 'linear-gradient(to right, rgba(255,255,255,1) 80%, rgba(255,255,255,0) 100%)',
```
iamstarkov: screen shot 2018-06-05 at 10 38 23 am

https://caniuse.com/#search=mask-image

iamstarkov:
also have you thought of using ready-to-use solution? like this one https://github.com/formidablelabs/nuka-carousel

src/components/carousel/carousel-styles.jsx
iamstarkov: @jesperwiner huh?

mjmonline:
@iamstarkov the Carousel is definitely replaceable. Tweaking it to replicate what we previously had with the custom Carousel will take a little time, not allot but worth mentioning. The size difference is about 20kb vs 60kb, not that huge difference. But also if we are keeping the same look and feel with the new Carousel on the discovery page then I don't know if its worth replacing it. I also definitely see the pros and that's why I'm torn and not sure how to proceed. What is your suggestion?

PR #401, opened 2018-06-05
fix(SegmentedControl): Fix race condition in onChange handler

bstream:
Also upgrade React to 16 in the dev dependencies


bstream:
Seems like Prettier wanted to play as well, thats why the formatting changed in one of the test files

PR #402, opened 2018-06-12
Feat/tooltip desktop only

JacobBlomgren:
A prop to display the tooltip popup only on desktop. A more complicated implementation is already used internally to achieve this functionality.

### DIFF
--- a/src/components/tooltip/tooltip-styles.jsx
+++ b/src/components/tooltip/tooltip-styles.jsx
@@ -53,6 +53,21 @@ export default theme => {
       },
     },
 
+    desktopOnlyWrapper: {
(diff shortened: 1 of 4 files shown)


src/components/tooltip/tooltip-styles.jsx
```diff
@@ -150,5 +150,11 @@ export default theme => {
       wordBreak: 'break-word',
       wordWrap: 'break-word',
     },
+    popupDesktopOnly: {
+      display: 'none',
+      [mixins.media('md')]: {
```
bstream: You can use `maxMedia` here instead so that you don't need to reset the `display`

PR #408, opened 2018-06-28
Feature/input editable

KitsuneKelso:
Adds a basic editable input type:

### DIFF
--- a/src/components/editable-input/editable-input.jsx
+++ b/src/components/editable-input/editable-input.jsx
@@ -1,0 +1,171 @@
+import React from 'react';
+import PropTypes from 'prop-types';
+import injectSheet from 'react-jss';
+import classNames from 'classnames';
+
+import styles from './editable-input-styles';
+import { Button, Icon, Input, Spinner } from '../../';
+
+class EditableInput extends React.Component {
+  constructor(props) {
+    super(props);
+
+    this.state = {
+      editing: false,
+      saving: false,
+      value: this.props.value,
+      originalValue: '',
+    };
+  }
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.value) this.setState({ value: nextProps.value });
+  }
+
+  onEdit = () => {
+    this.setState({ originalValue: this.state.value, editing: true });
+  };
+
+  onFocus = event => {
+    if (this.props.onFocus) {
+      this.props.onFocus(event);
+    }
+  };
+
+  onBlur = event => {
+    if (this.props.onBlur) {
+      this.props.onBlur(event);
+    }
+  };
+
+  onChange = event => {
+    this.setState({ value: event.target.value || this.props.emptyDefaultValue });
+
+    if (this.props.onChange) {
+      this.props.onChange(event);
+    }
+  };
+
+  onSubmit = async event => {
+    event.preventDefault();
+    if (!this.props.hasError) {
+      if (this.props.onSubmit) {
+        this.setState({ saving: true });
+        await this.props.onSubmit(this.state.value);
+        this.setState({ saving: false });
+      }
+      this.setState({ editing: false });
+    }
+  };
+
+  onCancel = () => {
+    this.setState({ value: this.state.originalValue });
+    if (this.props.onCancel) {
+      this.props.onCancel(this.state.originalValue);
+    }
+    this.setState({ editing: false });
+  };
+
+  render() {
+    const { classes } = this.props;
+
+    const input = this.state.editing ? (
+      <form className={classes.form} onSubmit={this.onSubmit}>
+        <Input
+          autoFocus
+          className={classes.input}
+          style={this.props.style}
+          type="text"
+          variant={this.props.variant}
+          disabled={this.props.disabled || this.state.saving}
+          label={this.props.label}
+          placeholder={this.props.placeholder}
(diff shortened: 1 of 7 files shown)


src/components/input/input-editable.jsx
```diff
@@ -0,0 +1,158 @@
...
+  onSubmit = async event => {
+    event.preventDefault();
+    if (!this.props.hasError) {
+      if (this.props.onSubmit) {
+        this.setState({ saving: true });
+        await this.props.onSubmit(this.state.value);
+        this.setState({ saving: false });
+      }
+      this.setState({ editing: !this.state.editing });
+    }
```
Dhq: Maybe safest to always set it to false here? I don't see any case where we want it to be true in this case (?)

src/components/input/input-editable.jsx
```diff
@@ -0,0 +1,158 @@
...
+      </div>
+    ) : (
+      <Button className={classes.buttonEdit} onClick={this.onEdit} variant="secondary" icon={<Icon.Edit />} />
+    );
+
+    return (
+      <div className={classNames(classes.inputEditable, this.props.className)}>
+        {renderInput}
+        {renderButtons}
+      </div>
```
Dhq: nitpick: just call them "input" and "buttons" or something instead? prefixing "render" implies a function to be ran in my mind

src/components/input/input.md
```diff
@@ -246,3 +246,14 @@ Type: `textarea`
...
       />
     </div>
+
+Type: `editable`
+
+    <div>
+      <span>What is your favorite color?</span>
+      <Input
+        label="My favorite color is"
+        value="Blue"
```
Dhq: Did you just assume my favourite color??

KitsuneKelso: Isn't blue everyone's favourite? 🤔

src/components/input/input-editable.jsx
KitsuneKelso: Good suggestion, will fix!

src/components/input/input-editable.jsx
```diff
@@ -0,0 +1,158 @@
...
+    );
+  }
+}
+
+InputEditable.propTypes = {
+  className: PropTypes.string,
```
JCB-K: could you extract these into a separate module and import them in both Input and InputEditable? Since I guess they're 95% the same

KitsuneKelso: On suggestion from @robineng I extracted the whole thing into a completely new component instead. :)

src/components/input/input-editable.jsx
KitsuneKelso: Done! :)

src/components/editable-input/editable-input.jsx
```diff
@@ -69,13 +69,14 @@ class InputEditable extends React.Component {
...
     const { classes } = this.props;
 
-    const renderInput = this.state.editing ? (
+    const input = this.state.editing ? (
       <form className={classes.form} onSubmit={this.onSubmit}>
         <Input
           autoFocus
           className={classes.input}
           style={this.props.style}
-          variant={this.props.variant === 'primary' ? 'secondary' : 'primary'}
```
mlenser: Is the default set elsewhere? Otherwise this is a breaking change and we need to verify every place where this is used.

src/components/editable-input/editable-input.jsx
```diff
@@ -99,29 +100,40 @@ class InputEditable extends React.Component {
...
+        <Button
+          variant="primary"
+          modifier="action"
+          onClick={this.onSubmit}
+          disabled={this.props.disabled || this.props.hasError || this.state.saving}
+        >
           {this.state.saving ? <Spinner size={13} color="#FFF" /> : this.props.submitLabel}
         </Button>
-        <Button variant="secondary" modifier="action" onClick={this.onCancel} disabled={this.state.saving}>
+        <Button variant="secondary" modifier="action" onClick={this.onCancel} disabled={this.props.disabled || this.state.saving}>
```
mlenser: Nit pick: If you're going to line break the primary (which I love), why not line break this one as well? Far more readable imo.

src/components/editable-input/editable-input.jsx
KitsuneKelso: It is set in the defaultProps on line 164 now instead. :)
```
EditableInput.defaultProps = {
  variant: 'secondary',
```

KitsuneKelso: Essentially, the way it was written until I broke out this into a completely new component was a workaround for the fact that `variant` was always set to `"primary"` as default for the regular input type. Now that it is it's own component, I can set the `defaultProps` the way I want them right away. 👍

src/components/editable-input/editable-input.jsx
KitsuneKelso: I agree completely, but I have not added the line breaks manually, I'm letting prettier handle that. It's a bit like the old problem we discussed previously, where prettier should be smarter about how it line breaks items such as this. I would much prefer if each value was on a separate line as well. I heard some talk about that prettier supposedly is getting smarter about this in newer versions, but since we're still stuck on the versions we are, we might have to live with this until we upgrade it.

PR #409, opened 2018-07-02
feat: Make PopupMenu accessible

JacobBlomgren:
Makes the popup menu work with tabs and arrow keys.

### DIFF
--- a/src/components/popup-menu/popup-menu.jsx
+++ b/src/components/popup-menu/popup-menu.jsx
@@ -6,10 +6,13 @@ import { Icon } from '../../';
 import keyCodes from './keyCodes';
 import PopupMenuList from './popup-menu-list';
 
+let idIncrement = 0;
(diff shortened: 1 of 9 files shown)


esundqvist:
Did some QA seems to work fine!

src/components/popup-menu/popup-menu.jsx
```diff
@@ -1,94 +1,145 @@
...
+    setTimeout(() => {
+      this.setState({ hasFocus: false });
+    }, 0);
+
+  onToggle = () => {
+    const { isOpen } = this.state;
+    if (!isOpen) {
+      document.addEventListener('click', this.documentClickOutsideListener);
+    } else {
+      document.removeEventListener('click', this.documentClickOutsideListener);
```
bstream: This should probably be called in componentWillUnmount as well to prevent possible memory leak

PR #411, opened 2018-07-04
fix(editable-input): helpText support and styling

KitsuneKelso:


src/components/editable-input/editable-input.jsx
```diff
@@ -136,6 +151,8 @@ class EditableInput extends React.Component {
 EditableInput.propTypes = {
   className: PropTypes.string,
   style: PropTypes.object,
+  /** Needs to be a valid input type */
+  type: PropTypes.string,
```
Dhq: Maybe use:
`type: PropTypes.oneOf(['text', 'tel', 'number', 'email']),`

KitsuneKelso: I just picked up the same propTypes validation as was used in the regular input for this prop, will probably leave as is for consistency.

src/components/editable-input/editable-input.md
```diff
@@ -18,3 +18,15 @@ With `emptyDefaultValue`
...
+
+With `hasError` and `helpText`
+
+    <div>
+    <span>Something is wrong here:</span>
+    <EditableInput
+            label="Is this correct?"
+            value="No"
+            hasError
+            helpText="It never is"
```
bstream: 😂

src/components/editable-input/editable-input.jsx
```diff
@@ -66,6 +67,16 @@ class EditableInput extends React.Component {
...
   };
 
+  renderHelpText() {
+    const modifiers = {
+      hasSuccess: this.props.hasSuccess,
+      hasWarning: this.props.hasWarning,
+      hasError: this.props.hasError,
+    };
+
+    return !this.props.helpText ? null : <HelpText {...modifiers}>{this.props.helpText}</HelpText>;
```
bstream: This logic might be better to put inside the `<HelpText />` component (to avoid this extra function)

PR #414, opened 2018-07-06
Revert football

FilipStenbeck:
Don´t merge as long as Sweden is in the worldcup


KitsuneKelso:
Guess it's time to remove it. 😢

PR #416, opened 2018-07-09
feat(tooltip): Add mobilePlacement prop to tooltip

JacobBlomgren:
Adds an optional mobile placement prop to the tooltip to prevent horizontal scrolling which has often
happened with horizontal placements. It's actually a (internally) somewhat messy workaround, needed because the smart placement is, and has been, broken for a long time. But the team decided that this was the best solution for now.


src/components/tooltip/tooltip-styles.jsx
```diff
@@ -69,70 +150,16 @@ export default theme => {
...
+      ...popupPlacementStyles(palette),
+      ...wrapInMobileMediaQuery(mixins, popupPlacementStyles(palette, 'Mobile')),
+      ...wrapInDesktopMediaQuery(mixins, popupPlacementStyles(palette, 'Desktop')),
     },
     popupFixed: {
       whiteSpace: 'inherit',
     },
 
     popupContent: {
+      position: 'static',
```
iamstarkov: it is the default one, isn't it?

JacobBlomgren: You're right, I moved out an inline style without thinking about it.

PR #417, opened 2018-07-09
feat(dropdown): add closeOnAction prop

JCB-K:
Add a prop that allows dropdowns to be closed on action clicks


src/components/dropdown/dropdown.jsx
```diff
@@ -108,10 +119,12 @@ Dropdown.propTypes = {
     }),
   ),
   actionsOpen: PropTypes.bool,
+  closeOnAction: PropTypes.bool,
 };
 
 Dropdown.defaultProps = {
   actionsOpen: false,
+  default: false,
```
KitsuneKelso: Was this supposed to be `closeOnAction: false,`?

Dhq: This is supposed to be: "closeOnAction" and not "default", I assume? :)

JCB-K:
@KitsuneKelso @Dhq I'm suffering of a bad case of the mondays, thanks for your consideration. Amended my commit!

PR #420, opened 2018-07-19
feat(button): support icon above the text

mlenser:
Icons currently are only supported to the left of text. This adds support to add an icon above text. Matches the behavior of the modal:

screen shot 2018-07-19 at 11 38 16

Example:

screen shot 2018-07-19 at 11 32 42


iamstarkov:
@mlenser dont forget to delete a branch after merge

PR #422, opened 2018-07-19
Fix/clean button spacing and add examples

mlenser:
Follow up to #420 
Had some local spacing that I missed and I've also added examples as punishment for my crime.

screen shot 2018-07-19 at 13 10 31


mlenser:
Can't rebase effectively. Closing and reopening.

PR #423, opened 2018-07-19
fix(button): button spacing and add examples

mlenser:
Follow up to #420 
Had some local spacing that I missed and I've also added examples as punishment for my crime.

screen shot 2018-07-19 at 13 10 31


iamstarkov:
is vertical alignment a bit off?

mlenser:
@iamstarkov do you mean compared to the other examples? This is taller and the examples do not have any flex wrapper that vertically aligns the elements.

PR #434, opened 2018-09-11
feat: Add pagination component

mjmonline:


src/components/pagination/page/page-styles.jsx
```diff
@@ -0,0 +1,43 @@
...
+    alignItems: 'center',
+    [mixins.media('md')]: {
+      display: 'flex',
+    },
+  },
+  itemAlwaysVisible: {
+    display: 'flex',
+  },
+  button: {
+    fontFamily: typography.primary.fontFamily,
```
bstream: You can also spread `typography.primary` here to also get `fontWeight`

src/components/pagination/page/page.jsx
```diff
@@ -0,0 +1,40 @@
...
+    className={cn(classes.item, {
+      [classes.itemAlwaysVisible]: isSelected || isFirst || isLast,
+    })}
+  >
+    <button
+      type="button"
+      className={cn(classes.button, {
+        [classes.buttonSelected]: isSelected,
+      })}
+      onClick={() => {
```
bstream: This function (eg. `() => { clickHandler(pageNumber) }`) should be a class-method instead (requires `Page` to be a class) so that it is not re-created on each render

src/components/pagination/pagination.jsx
```diff
@@ -0,0 +1,172 @@
+import React, { Component } from 'react';
+import PropTypes from 'prop-types';
+import injectSheet from 'react-jss';
+import { Icon } from '../../';
+import Page from './page';
+import Stepper from './stepper';
+import Ellipsis from './ellipsis';
+import styles from './pagination-styles';
+
+const getRandomString = () =>
```
bstream: This will break on SSR (Math.random() will evaluate differently on server and client). Can we make this deterministic somehow?

src/components/pagination/pagination.jsx
```diff
@@ -0,0 +1,172 @@
...
+    } else {
+      /*
+      Add pages to Pagination
+        * Anchors (far left and far right)
+        * Selected item
+        * Siblings to the selected item
+        * Lastly fills out with pages until no empty spots left
+      */
+
+      for (let i = 1; i <= this.pagesCount; i += 1) {
```
bstream: This can probably be done with a map to keep some functional style :)

src/components/pagination/pagination.jsx
```diff
@@ -0,0 +1,172 @@
...
+    super(props);
+    const { total, limit, selected, anchors, selectedSiblings } = this.props;
+
+    this.state = {
+      selected: parseInt(selected, 10),
+    };
+
+    this.pagesCount = Math.ceil(total / limit);
+    this.pagesToDraw = 1 + anchors * 2 + selectedSiblings * 2; // 1 = the selected item
+    this.advancedPagination = this.pagesCount > this.pagesToDraw;
```
bstream: Are these (pagesCount, pagesToDraw, advancedPagination) directly or indirectly being used in the render cycle? Then they should probably be in the state instead. 

Also, since they currently rely on props, they should probably be re-calculated if/when props change.

src/components/pagination/pagination.jsx
```diff
@@ -0,0 +1,172 @@
...
+  handlePageSelected = number => {
+    const { changeHandler } = this.props;
+    this.setState({
+      selected: number,
+    });
+
+    changeHandler(number);
+  };
+
+  paginate() {
```
bstream: This is a method that from props and state render components, maybe it can be extracted to a component?

src/components/pagination/pagination.jsx
```diff
@@ -0,0 +1,172 @@
...
+    };
+
+    this.pagesCount = Math.ceil(total / limit);
+    this.pagesToDraw = 1 + anchors * 2 + selectedSiblings * 2; // 1 = the selected item
+    this.advancedPagination = this.pagesCount > this.pagesToDraw;
+  }
+
+  getEllipsis = hiddenOnDesktop => <Ellipsis key={getRandomString()} hiddenOnDesktop={hiddenOnDesktop || !this.advancedPagination} />;
+
+  getPage(pageNumber) {
```
bstream: If possible avoid this render-method approach, either inline the body or make this a separate component imo.

mjmonline:
@bstream please have another look at this

src/components/pagination/pagination.jsx
```diff
@@ -80,8 +91,14 @@ Pagination.propTypes = {
   total: PropTypes.number.isRequired,
   limit: PropTypes.number,
   changeHandler: PropTypes.func,
+  /** Screenreader friendly title for the pagination */
+  title: PropTypes.string,
   buttonTextPrevious: PropTypes.string,
   buttonTextNext: PropTypes.string,
+  /** A page number will be appended to this string */
+  pageLabelText: PropTypes.string,
```
JacobBlomgren: we have a bunch of these now, should they be grouped in one object?

PR #436, opened 2018-09-21
feat(EditableInput): Add onEdit prop to run when toggling edit

Dhq:


src/components/editable-input/editable-input.jsx
```diff
@@ -25,6 +25,10 @@ class EditableInput extends React.Component {
 
   onEdit = () => {
     this.setState({ originalValue: this.state.value, editing: true });
+
+    if (this.props.onEdit) {
+      this.props.onEdit();
```
JCB-K: you might want to pass the onEdit event to the function? like we do with onChange, onSubmit. Not sure if its required, but at least its consistent.

Dhq: That is very true! Added it now :)

PR #438, opened 2018-10-01
A11y fixes

bstream:


src/components/labeled-value/labeled-value.jsx
```diff
@@ -80,14 +90,14 @@ function LabeledValue({
   const labelClasses = cn(classes.label);
   const valueClasses = cn(classes.value, classes[`value-${size}`]);
 
+  const renderChild = setIdOnChild(id);
+
   return (
     <div {...rest} className={rootClasses}>
-      <span className={labelClasses} id={id}>
+      <label className={labelClasses} htmlFor={id}>
```
mjmonline: I'm not sure it's valid to use `<label>` if not together with a `<input />`

bstream: Good point! We probably need to re-think this some more... :)

JacobBlomgren: Yeah, the problem is that it most often used together with our `<Input />` that ends up rendering a `<input />`

bstream:
This needs to be thought of some more, closing for now!

PR #450, opened 2018-10-16
feat: add tabs component

mjmonline:


src/components/tabs/tabs.md
```diff
@@ -0,0 +1,94 @@
...
+                <Tabpanel>
+                  Content for panel: {this.state.value}
+                </Tabpanel>
+              </Tabs>
+            </div>
+          );
+        }
+      }
+
+      <TabsExample3 />
```
Unic0arn: An example of singlePanel could be interesting

mjmonline: This is the example of singlePanel :)

src/components/tab/tab-styles.jsx
```diff
@@ -0,0 +1,106 @@
...
+
+      '&:last-child': {
+        paddingRight: 0,
+      },
+    },
+
+    tab: {
+      position: 'relative',
+      display: 'block',
+      fontSize: '12px',
```
jesperwiner: 12

src/components/tab/tab-styles.jsx
```diff
@@ -0,0 +1,106 @@
...
+      fontFamily: typography.primary.fontFamily,
+      fontWeight: typography.fontWeightSemiBold,
+      paddingTop: 5,
+      paddingBottom: borderHeight + 2,
+      textTransform: 'uppercase',
+      letterSpacing: 0.5,
+      whiteSpace: 'nowrap',
+      cursor: 'pointer',
+
+      'button&': {
```
jesperwiner: whats these with & at end ?

src/components/tab/tab-styles.jsx
```diff
@@ -0,0 +1,106 @@
...
+        background: 'none',
+        border: 0,
+      },
+
+      'a&': {
```
jesperwiner: same here

src/components/tabpanel/tabpanel-styles.jsx
```diff
@@ -0,0 +1,9 @@
+export default () => ({
+  root: {
+    padding: [20, 0],
+
+    '&[hidden]': {
```
jesperwiner: how does this selector work ?

mjmonline: this selector says if root element has the attribute hidden="true" or just hidden then these rules should be applied

src/components/tab/tab-styles.jsx
mjmonline: & is the parent selector, in this case we are selecting all a tags that also has the parent class on it. Same for your button comment :)

src/components/tabpanel/tabpanel-styles.jsx
```diff
@@ -0,0 +1,9 @@
+export default () => ({
```
KitsuneKelso: If you're not using anything from the theme, this could just be simplified to:
```
export default {
 // styles here
};

File: src/components/tabs/tabs-styles.jsx [hunk 7]
```diff
@@ -0,0 +1,18 @@
+export default theme => {
+  const { palette } = theme;
```
KitsuneKelso: Same here as on the other stylesheet:
```
export default ({ palette }) => ({
```

src/components/tabs/tabs.jsx
```diff
@@ -0,0 +1,66 @@
...
+
+      index += 1;
+
+      return childrenWithContext;
+    });
+  }
+
+  render() {
+    const { classes, className, variant } = this.props;
+    const tabs = this.injectContextProps('Jss(Tab)');
```
wemstad: Do we really want to do this? Feels like there should be a better solution that to connect to 'Jss(Tab)'

PR #453, opened 2018-10-17
fix: add type to button

mjmonline:
this prevents form to submit on click by default

### DIFF
--- a/src/components/popup-menu/popup-menu-item.jsx
+++ b/src/components/popup-menu/popup-menu-item.jsx
@@ -6,6 +6,8 @@ import styles from './popup-menu.styles';
 
 const PopupMenuItem = ({ node, children, topBorder, linkTo, onClick, onKeyDown, listItemRef, disabled, classes, onFocus, className }) => {
   const Element = disabled ? 'button' : node;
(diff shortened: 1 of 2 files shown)


src/components/popup-menu/popup-menu-item.jsx
```diff
@@ -15,7 +15,15 @@ const PopupMenuItem = ({ node, children, topBorder, linkTo, onClick, onKeyDown,
         * call focus on the ref from the PopupMenuList.
       */}
       <span ref={ref => ref && listItemRef && listItemRef(ref.childNodes[0])}>
-        <Element to={linkTo} onClick={onClick} onFocus={onFocus} onKeyDown={onKeyDown} className={classes.link} disabled={disabled}>
+        <Element
+          type={Element === 'button' ? 'button' : null}
```
jesperwiner: type=null is valid ? or just ignored then ?

mjmonline: yes it will be ignored

mjmonline: Do we prefer to set it to `false` instead? That will aslo be ignored

PR #456, opened 2018-10-19
feat: add url generator prop

mjmonline:


mjmonline:
A question before merging, should we keep the old behavior of the pagination? The old behavior was that each page in the pagination is a `<button>` which triggers a event handler. Whit this new `urlGenerator` prop we have react-router creating proper `<a>` tags. This also means that this component could be stateless

src/components/pagination/page/page.jsx
```diff
@@ -2,6 +2,7 @@ import React, { Component } from 'react';
 import PropTypes from 'prop-types';
 import injectSheet from 'react-jss';
 import cn from 'classnames';
+import { Link } from 'react-router-dom';
```
bstream: This is not a current dependency, is it possible to send it in as a prop instead similar to how `Button` works?

iamstarkov: agreed, it shouldn't be here

src/components/pagination/page/page.jsx
```diff
@@ -12,24 +13,27 @@ class Page extends Component {
   };
 
   render() {
-    const { classes, isSelected, isFirst, isLast, labelText, children } = this.props;
+    const { classes, isSelected, isFirst, isLast, labelText, url, children } = this.props;
+    const disabled = isSelected;
+    const Element = url && !disabled ? Link : 'button';
```
iamstarkov: ask for element, default it to 'button'

src/components/pagination/page/page.jsx
```diff
@@ -12,24 +13,27 @@ class Page extends Component {
...
+    const customProps = Element === Link ? { to: url } : { type: 'button', disabled };
```
iamstarkov: `{ to: url }` is a implementation detail for RR, shouldnt be here

src/components/pagination/pagination.md
```diff
@@ -1,31 +1,70 @@
...
-    <div>
-      <Pagination selected={2} total={100} changeHandler={() => {}} />
-    </div>
+      urlGenerator(pageNumber) {
+        return `?page=${pageNumber}`
+      }
+
+      render() {
+        return (
+          <HashRouter>
```
iamstarkov: have you tried `MemoryRouter` instead? I believe Hash based router will conflict with styleguidist

src/components/pagination/page/page.jsx
```diff
@@ -12,24 +13,27 @@ class Page extends Component {
...
           className={cn(classes.button, {
             [classes.buttonSelected]: isSelected,
           })}
           onClick={this.onPageClick}
-          disabled={isSelected}
           aria-label={labelText}
         >
           {children}
-        </button>
+        </Element>
```
iamstarkov: what do you think of making internal Element to be controllable from the outside? like `renderItem` props which will get all relevant data from this component to be able relevant Element?

iamstarkov:
@mjmonline we should aim for backwards compatibility, so we should introduce new behaviour while keeping previous contract in place

iamstarkov:
if it is too hard, consider brand new component

src/components/pagination/page/page.jsx
KitsuneKelso: Sounds like a good idea!

mjmonline: @iamstarkov Not sure what you mean

src/components/pagination/page/page-styles.jsx
```diff
@@ -12,24 +15,31 @@ export default ({ palette, typography, mixins }) => ({
...
+    backgroundColor: palette.color.white,
     fontSize: 16,
     padding: 0,
-    width: 45,
-    height: 45,
-    border: `2px solid ${palette.action.disabled}`,
+    width: buttonSize,
+    height: buttonSize,
+    lineHeight: `${buttonSize - buttonBorderSize * 2}px`,
+    border: `${buttonBorderSize}px solid ${palette.action.disabled}`,
```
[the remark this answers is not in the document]
mjmonline: I prefer to keep it as close to css as possible

bstream:
Well done @mjmonline! 🎉

PR #464, opened 2018-11-05
feat(Tabs): add className prop

robineng:


src/components/tabpanel/tabpanel-styles.jsx
```diff
@@ -1,7 +1,5 @@
 export default {
   root: {
-    padding: [20, 0],
```
robineng: This is a breaking change but it's better to do this now than later

Unic0arn: ping @mjmonline , I think it's better if we add this padding in our component instead

bstream: It can break visually, but the API of the component has not changed so it should be ok :)

mjmonline: Looks good!

PR #465, opened 2018-11-07
fix: resolve styling bug in safari

wemstad:
Safari has a quirk when rendering transparent in linear-gradients causing the gradient to darken instead. Using rgba resolves that issue.
Before: 
before
After:
after

Verified that it looks good in: 
Safari 12
Chrome 70
Firefox 63
IE 11
Edge 42


robineng:
👌

PR #466, opened 2018-11-07
Fix flag subpixel coloring & deprecated flag dimensions

bstream:


src/components/flag/flag-deprecated.jsx
```diff
@@ -5,43 +5,43 @@ import injectSheet from 'react-jss';
...
 
-const addBorder = props => !props.hideBorder && !props.round && !props.secondaryCountryCode;
+export const styles = () => ({
+  container: {
+    display: 'inline-block',
+  },
+  flagStyle: {
+    display: 'block',
+    boxSizing: 'border-box',
+    left: 0,
```
mjmonline: Is this needed? It feels like this together with `position: relative` will move the flag including the border `0px` to the left.

PR #473, opened 2018-11-30
Empty table headers

wemstad:
## Issue
In IE and Edge the header border on the tables does not render correctly if the `<th>` is empty. 
Expected:
Expected
Actual:
Actual

## Background 
This is caused by a parent component having `display: flex` that Edge picks up and incorrectly applies it to the` div` in the `th`. I have not been able to override this effect. 

## Fix
The fix in this PR injects a white space in any table cell that would otherwise be empty. Hence, render the table correctly.

### DIFF
--- a/package.json
+++ b/package.json
@@ -114,7 +114,7 @@
     "in-publish": "^2.0.0",
     "istanbul-instrumenter-loader": "^2.0.0",
     "jsdom": "^9.11.0",
-    "lint-staged": "^7.0.0",
+    "lint-staged": "^7.2.0",
     "lodash.camelcase": "^4.3.0",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -114,7 +114,7 @@
     "in-publish": "^2.0.0",
     "istanbul-instrumenter-loader": "^2.0.0",
     "jsdom": "^9.11.0",
-    "lint-staged": "^7.0.0",
```
bstream: 🤔 this did not change the resolved version, so maybe not needed? :)

wemstad: Interesting. The reason why I updated it was that I had issues with running it with node 10. Should not impact the code in any way.

PR #485, opened 2018-12-19
fix(TableContentWrapper): add ability to set thProps on cols

Unic0arn:


src/components/table-content-wrapper/table-head/table-head.jsx
```diff
@@ -113,6 +120,8 @@ class TableHead extends Component {
                 scope="col"
                 role="columnheader"
                 unstyledChild
+                valign="bottom"
```
jesperwiner: isn't valign like obsolete and shouldn't be used ?

mjmonline: true, this is deprecated. Sorry @Unic0arn  :)

KitsuneKelso: To elaborate, you should get the behaviour you're aiming for with css instead:
```
vertical-align: bottom;
```

PR #488, opened 2018-12-20
fix: make caption visible if wanted

mjmonline:


src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -47,7 +49,9 @@ class TableContentWrapper extends Component {
       return (
         <div className={classes.root}>
           <Table tableLayoutFixed>
-            <caption className={classes.screenReadersOnly}>{localization.caption}</caption>
+            <caption className={cn(classes.caption, 'smart-table__caption', { [classes.screenReadersOnly]: captionHidden })}>
```
KitsuneKelso: Why the hardcoded class `smart-table__caption`?

Dhq: "smart-table__caption" class should probably not always be applied here right?

mjmonline: Correct, should not always be there

mjmonline: I want to target the caption from outside the component, but it shouldn't always be there

PR #491, opened 2018-12-21
fix: add ability to show caption on mobile or desktop only

mjmonline:


src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -21,7 +21,8 @@ class TableContentWrapper extends Component {
   render() {
     const {
       classes,
-      captionHidden,
```
JacobBlomgren: are we changing/breaking the api here?

src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -40,6 +41,8 @@ class TableContentWrapper extends Component {
       trProps,
       borderlessRows,
     } = this.props;
+    const captionHidden = !captionVisibleMobile && !captionVisibleDesktop;
+    const captionVisible = captionVisibleMobile || captionVisibleDesktop;
```
KitsuneKelso: `captionHidden` is a bit redundant here, it's the same as if `captionVisible` is false.

src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -21,7 +21,8 @@ class TableContentWrapper extends Component {
   render() {
     const {
       classes,
-      captionHidden,
+      captionVisibleMobile,
```
JacobBlomgren: Is there a possibility to send in `captionVisible = true` as shortcase, and the option for `captionVisible = { mobile: true }` if you want only one visible? That would result in less typing/props without that much extra complexity.

src/components/table-content-wrapper/table-content-wrapper.styles.js
```diff
@@ -7,6 +7,16 @@ export default ({ mixins, palette, typography }) => ({
     borderBottom: [2, 'solid', palette.color.grayDarker],
   },
   screenReadersOnly: mixins.screenReadersOnly,
+  captionMobileHidden: {
+    [mixins.maxMedia('md')]: {
+      ...mixins.screenReadersOnly,
```
JacobBlomgren: can we do `[mixins.maxMedia('md')]: mixins.screenReadersOnly` instead? Now we're only spreading one object

src/components/table-content-wrapper/table-content-wrapper.jsx
mjmonline: I added this yesterday night

src/components/table-content-wrapper/table-content-wrapper.styles.js
JacobBlomgren: Oh and btw nice a11y! 💪

src/components/table-content-wrapper/table-content-wrapper.jsx
JacobBlomgren: Can we in some way merge together my suggestion below and keeping `captionHidden`?

mjmonline: But why? No one has used this, it's only me working on this

src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -104,8 +100,13 @@ class TableContentWrapper extends Component {
 
 TableContentWrapper.propTypes = {
   classes: PropTypes.shape().isRequired,
-  captionVisibleMobile: PropTypes.bool,
-  captionVisibleDesktop: PropTypes.bool,
+  captionVisible: PropTypes.oneOfType([
+    PropTypes.bool,
+    PropTypes.shape({
```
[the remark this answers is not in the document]
mjmonline: It shouldn't be a problem because they're not required

src/components/table-content-wrapper/table-content-wrapper.jsx
```diff
@@ -21,6 +21,7 @@ class TableContentWrapper extends Component {
   render() {
     const {
       classes,
+      captionHidden,
```
JacobBlomgren: Wait don't we get both `captionHidden` and `captionVisible` now? Sorry for the inconvenience  😅

mjmonline: It should be fine as long as you dont use both

JacobBlomgren: but why do we need both then? can't we keep `captionVisible` and just add the object prop possibility?

mjmonline:
@iamstarkov @JacobBlomgren please check again

jesperwiner:
@JacobBlomgren how do we mark stuff as experimental ?
After the latest adjustments it's not breaking any more so approving it.
Can be that we redo this component once all the list pages settled in and we deprecate this one then,

JacobBlomgren:
We discussed it one more round and decided to not mark stuff as experimental for now but rather iron experimental components out internally before adding them to the ui kit. 

What would be nice though is to put all deprecated components at the bottom of the github.io page eventually, since we have started to deprecate a lot (which is good).

PR #497, opened 2019-01-14
fix(PopupMenu): add prop popupMenuDropdownZIndex

Unic0arn:
Allow specifying z-index on the dropdown from the popupMenu

### DIFF
--- a/src/components/popup-menu/popup-menu-list.jsx
+++ b/src/components/popup-menu/popup-menu-list.jsx
@@ -69,14 +69,24 @@ class PopupMenuList extends Component {
   };
 
   render() {
-    const { width, classes, children, enter, exit, maxHeight, isOpen, 'aria-labelledby': ariaLabelledBy } = this.props;
+    const {
+      width,
+      classes,
+      children,
+      enter,
+      exit,
+      maxHeight,
+      isOpen,
+      'aria-labelledby': ariaLabelledBy,
+      popupMenuDropdownZIndex,
+    } = this.props;
     const itemContainerStyle = maxHeight === 'none' ? {} : { maxHeight, overflowY: 'scroll' };
     let notDisabledIndex = 0;
     return (
       <TransitionGroup>
         {isOpen && (
           <CSSTransition classNames={classes.menuSlideDown} timeout={{ exit, enter }}>
-            <div className={classes.menuPopup} style={{ width }}>
+            <div className={classes.menuPopup} style={{ width, zIndex: popupMenuDropdownZIndex }}>
               <div className={classes.menuItemContainer} style={itemContainerStyle}>
                 <ul className={classes.menuItems} aria-labelledby={ariaLabelledBy} ref={this.setListElement} onBlur={this.onBlur}>
                   {Children.map(children, child => {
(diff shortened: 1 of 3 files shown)


JacobBlomgren:
What's the reason behind this?

JacobBlomgren:
> Sticky headers on tables also have z-index 1 so the popup renders behind the table header if that's enabled.

Ok. Could we just change the z-index here instead of passing it in as a prop? That feels like a solution that will not mature nicely, I'm afraid.

src/components/popup-menu/popup-menu-list.jsx
```diff
@@ -69,14 +69,24 @@ class PopupMenuList extends Component {
...
+    const {
+      width,
+      classes,
+      children,
+      enter,
+      exit,
+      maxHeight,
+      isOpen,
+      'aria-labelledby': ariaLabelledBy,
+      popupMenuDropdownZIndex,
```
JacobBlomgren: Either way I would prefer to have this named `zIndex`

Unic0arn: I prefer to have it explicit what I'm setting the z-index of

Unic0arn:
> > Sticky headers on tables also have z-index 1 so the popup renders behind the table header if that's enabled.
> 
> Ok. Could we just change the z-index here instead of passing it in as a prop? That feels like a solution that will not mature nicely, I'm afraid.

Why will it not mature nicely? I prefer to have it default to 1 and then if the need arises we have the possibility to override it. Also, adding this as a prop won't change current behaviour if someone relies on it being 1.

src/components/popup-menu/popup-menu-list.jsx
JacobBlomgren: what about `dropdownZIndex` then?

JacobBlomgren:
> Why will it not mature nicely? I prefer to have it default to 1 and then if the need arises we have the possibility to override it. Also, adding this as a prop won't change current behaviour if someone relies on it being 1.

Because I feel the problem we have is that we go about changing our components the wrong way, by customizing rather than considering if the component everywhere should be updated. Which makes all updates harder. 

If we allow this as a prop, there will be several hard to find places in the wild with different z-indeces. It's better to contain to as few places as possible. And also go over our current usage of them because I know that this component clashes with the mobile navbar on our site too.

Unic0arn:
> > Why will it not mature nicely? I prefer to have it default to 1 and then if the need arises we have the possibility to override it. Also, adding this as a prop won't change current behaviour if someone relies on it being 1.
> 
> Because I feel the problem we have is that we go about changing our components the wrong way, by customizing rather than considering if the component everywhere should be updated. Which makes all updates harder.
> 
> If we allow this as a prop, there will be several hard to find places in the wild with different z-indeces. It's better to contain it to as few places as possible. And also go over our current usage of them because I know that this component clashes with the mobile navbar on our site too.

I disagree. z-indices are hard enough to deal with without having arbitrary numbers specified deep in sub components. I think having default values of 1 like here and then in the special case where it's needed overwrite it is better. 

Bumping it to 2 here won't solve the bottom nav issue. That issue should be solved by looking at the stacking contexts of the menus.

JacobBlomgren:
> I disagree. z-indices are hard enough to deal with without having arbitrary numbers specified deep in sub components. I think having default values of 1 like here and then in the special case where it's needed overwrite it is better.
> 
> Bumping it to 2 here won't solve the bottom nav issue. That issue should be solved by looking at the stacking contexts of the menus.

No it doesn't solve it, but neither does allowing the possibility for even more spread out arbitrary numbers, imo.

PR #498, opened 2019-01-14
fix(Tab): expand to allow node (for example router Link)

jesperwiner:


src/components/tab/tab.jsx
```diff
@@ -56,6 +63,8 @@ Tab.propTypes = {
   singlePanel: PropTypes.bool,
   changeHandler: PropTypes.func,
   className: PropTypes.string,
+  node: PropTypes.oneOfType([PropTypes.string, PropTypes.func]),
+  to: PropTypes.string,
```
JacobBlomgren: Gonna want a default on this one

src/components/tab/tab.jsx
```diff
@@ -36,8 +42,9 @@ class Tab extends Component {
           href={href}
           role="tab"
           type={Element === 'button' ? 'button' : null}
+          onClick={node ? null : this.handleClick}
```
JacobBlomgren: maybe put this in `handleClick` instead, since it is "`handleClick`" logic.

src/components/tab/tab.jsx
```diff
@@ -36,8 +42,9 @@ class Tab extends Component {
           href={href}
```
sievertknorrblad: will they get the correct link on hover if href is set like this?

src/components/tab/tab.jsx
JacobBlomgren: From our eslint rules:

>> Always define explicit defaultProps for all non-required props.

>Why? propTypes are a form of documentation, and providing defaultProps means the reader of your code doesn’t have to assume as much. In addition, it can mean that your code can omit certain type checks.

https://github.com/airbnb/javascript/tree/master/react#props

JacobBlomgren: I think this rule came after the version we have in the ui kit as of now, so the linting didn't catch it, but the reason for having it still stands.

JacobBlomgren:
@mjmonline @jesperwiner and I talked more offline, hold on with merging for a while, while we sort it out.

PR #499, opened 2019-01-15
fix: add height 100% to th and th:after to fix border issue in edge

Unic0arn:
Fixing issue in edge where the bottom border on Th elements went all over the place.
screenshot 2019-01-15 at 11 42 17


wemstad:
LGTM

PR #500, opened 2019-01-15
Group together related components to same folders

JacobBlomgren:
Cleans up a lot of pollution in `./src`.

table-cell --> table/table-cell
tbody --> table/tbody
td --> table/td
tfoot --> table/tfoot
th --> table/th
thead --> table/thead
tr --> table/tr

tab --> tabs/tab
tabpanel --> tabs/tabpanel

Shouldn't be breaking since our only outwardly supported, and documented, API is to import like so: `import { Table } from 'nordnet-ui-kit`.


Dhq (approved):
🥳

jesperwiner (approved):

PR #504, opened 2019-01-16
fix: fix border bottom rendering on sticky headers in edge

Unic0arn:
Second attempt at this. 
Only setting height 100% on Ths now to avoid breaking responsive rendering
Switched to setting border on ::before to avoid it overlapping content when setting height 100% on that.


wemstad:
I ran the whole site locally with this change and cant find any issues with it. Feels like a to small change to require a full experimental build.

PR #506, opened 2019-01-17
chore: remove node 6 from testing env.

wemstad:


iamstarkov:
what's up with all the changes other than removal of node 6?

wemstad:
> what's up with all the changes other than removal of node 6?

Autoformatting, shouldn't be there. Will fix. 

> End of life for node 6 is April 2019, so it is a little bit early to drop support for it. but its okay

Yeah, but I don't really see any reason for us to continue supporting it. Bleeding egde ftw!

iamstarkov:
> Autoformatting, shouldn't be there. Will fix.

thanks

> Yeah, but I don't really see any reason for us to continue supporting it.

I agree

> Bleeding egde ftw!

I disagree

PR #507, opened 2019-01-17
fix: fix th covering whole page in IE11

robineng:


src/components/table/thead/thead-styles.jsx
```diff
@@ -22,6 +22,9 @@ const normal = ({ mixins, typography, palette: { color } }) => ({
     border: 0,
     '& th': {
       position: 'sticky',
+      fallbacks: {
```
iamstarkov: read more here https://cssinjs.org/jss-syntax?v=v10.0.0-alpha.8#fallbacks

PR #509, opened 2019-01-24
feat(dist): Provide both CJS and ESM in dist to enable tree shaking

Dhq:
Provide both CJS and ESM in dist to enable tree shaking, breaking change since we change the structure of the dist folder. Even though it looks like none of us imported anything dependent on the dist file structure :)


.babelrc
```diff
@@ -1,15 +1,35 @@
 {
```
Dhq: Some lovely day we will gather the power to upgrade to babel 7 so this config can be less verbose, today is not that day...

JacobBlomgren:
This is awesome, but if it's breaking I have a lot of stuff that could (should?) go into that release too

Dhq:
@JacobBlomgren Hmm yea, if it's something that can be merge-ready in a few days sure! Else I would argue that we merge this, and create another breaking for those changes.

iamstarkov:
we verified that path related breaking change won't require any change from internal consumers apart from version bump

JacobBlomgren:
@iamstarkov We talked about a v3 branch to pile up on. So you're say that it's not breaking or am I mistaken?

iamstarkov:
I say this particular PR is the breaking change *de jure*, but *de facto* is not because non of internal consumers will need to fix anything and we can safely bump to the new version in the main app

JacobBlomgren:
@iamstarkov ok so you just want to release this?

iamstarkov:
it is a possibility we should explore

JacobBlomgren:
We have discussed offline and will merge this temporarily into a v3 branch. @Dhq could you do this please?

Dhq:
@JacobBlomgren yep, done!

PR #511, opened 2019-01-29
Removed deprecated components and props

JacobBlomgren:
@jauniuse will complement this PR with a fix to the `Flag` component


bstream:
Here are some things that we might want to consider for this issue:

* Deprecate `Dropdown` and some of the icons? (or maybe remove them for this version)?
* `CostInfoLogo` is quite odd, maybe deprecate / remove?
* `Flag`- Literal flag size should maybe also be deprecated in favour of size?
* `InstrumentBadge` - internalize? don't know how useful it is here.
* `Subsection` - this is currently checking window width and should probably not in the future (since that creates CSR-SSR mismatch)

JacobBlomgren:
@bstream Nice. I already had `Dropdown` in mind. I think for a smooth transition it should be marked as deprecated with some breathing room before being removed. However, not used internally so idk. 

Jaunius is onto Flag deprecations.

I/we also thought about deprecating `TableContentWrapper` and `Tabs`. And on `Table` maybe we can deprecate `colorAlternateRows` which isn't used anywhere, as @wemstad has noted? Is there more that can be deprecated from `Table` @jesperwiner? We don't need to worry about removing it, just marking it as deprecated for now.

bstream:
> @bstream Nice. I already had `Dropdown` in mind. I think for a smooth transition it should be marked as deprecated with some breathing room before being removed. However, not used internally so idk.

Maybe we could mark them as deprecated in another branch and merge that first? (eg. deprecate them in the last 2.x version. And then remove in 3.x)

iamstarkov:
any migration guide updates?

PR #516, opened 2019-01-31
Upgrade babel, eslint, webpack, styleguidist and more 

JacobBlomgren:
Upgraded almost all devDependencies except for eslint, where I upgraded eslint one major version but left the eslint plugins untouched, and got rid of mocha-webpack. All should stay the same.


src/components/alert/alert.md
```diff
@@ -1,8 +1,10 @@
-    <div>
```
JacobBlomgren: All of these `.md` files are changed because with the new styleguidist, you have to wrap code examples in
\```js
\<Alert />
\```. 

That's the only thing that changed

Dhq:
Ahhh, freshness!

iamstarkov:
@JacobBlomgren dont forget to delete branches

PR #519, opened 2019-02-01
ci(Travis): avoid running tests twice

### LINKED ISSUE #518: Avoid running tests twice in CI
@Dhq and I talked, and travis currently first runs tests, then runs tests again as part of coveralls. I propose that we let `test` be, for local development, but run it once with `nyc` in CI. That way we only need to run it once.

JacobBlomgren:
Closes #518


JacobBlomgren:
I'm closing this since I didn't get it to work. We're going to turn off building only on PRs to avoid spamming PRs when working on the Travis.

JacobBlomgren:
Also, we're considering investigating a codemod to change to jest or ava, and then this would be superfluous.

PR #521, opened 2019-02-01
Change mocha to jest 

### LINKED ISSUE #520: Investigate possibility of converting test framework to jest

JacobBlomgren:
Closes #520


.eslintrc
```diff
@@ -14,19 +15,17 @@
...
     "browser": true,
-    "es6": true,
-    "mocha": true
+    "es6": true
   },
   "rules": {
     "import/no-extraneous-dependencies": 0,
     "react/forbid-prop-types": [2, { "forbid": ["any", "array"] }],
     "react/require-default-props": 0,
-    "mocha/no-exclusive-tests": 2,
```
iamstarkov: any similar rules for jest?

JacobBlomgren: Yeah we got a bunch from the added jest plugin, in fact even more 🎉

.travis.yml
```diff
@@ -16,6 +13,12 @@ cache:
 
 jobs:
   include:
+    - stage: coverage
+      node_js:
+        - 10
+        - 8
```
iamstarkov: why do you want to run coverage twice?

src/components/pane/pane.test.js
```diff
@@ -41,39 +40,28 @@ describe('<Pane />', () => {
...
+      expect(renderedTabs.at(0).hasClass(classes.active)).toBe(true);
     });
 
-    it('default size should be md', () => {
-      expect(renderedTabs.at(0).hasClass(classes.md)).to.equal(true);
-    });
-
-    it('should change active tab on activeTab prop change', () => {
-      wrapper.setProps({ activeTab: 1 });
-      expect(renderedTabs.at(1).hasClass(classes.active));
```
JacobBlomgren: The eslint jest plugin caught that these two never had any assertions. Currently they both fail, if I make the assumption that the author of the test wanted this to be `true`. Since this is a deprecated component, I decided not to fix these two.

.travis.yml
JacobBlomgren: We currently do already, see #508. It seems as if coveralls recently has become smart enough to ignore the duplicate (check more recent PRs)

JacobBlomgren: In the future I could envision being smart and trying to avoid this, but I prefer keeping the PRs modular and do the jest migration first.

iamstarkov: so like no travis config change?

JacobBlomgren: I changed travis because jest was slightly slower (it can be a number of reasons, really), so I wanted to avoid what we previously did, which was run tests, then run tests _again_ for coveralls.

iamstarkov: hm. I believe jest generate coverage report without running tests if you ran tests before

JacobBlomgren: Yeah, I tried this with mocha before switching to jest, but I couldn't seem to access the coverage file then from the `after_script` or `after_success`, where coveralls was run. 🤔

PR #523, opened 2019-02-05
feat(icons): add co2 icon

anncos:


src/components/icon/icons/co2.jsx
```diff
@@ -0,0 +1,35 @@
...
+  strokeWidth,
+  style: styleProp,
+  ...rest // eslint-disable-line comma-dangle
+}) {
+  const style = {
+    ...styleProp,
+    ...Icon.defaultProps.style,
+  };
+  return (
+    <svg style={style} {...rest}>
```
JacobBlomgren: There should be a better way to do this in the future instead of 35 copy pasted lines of code

anncos: Ok :)

PR #527, opened 2019-02-08
feat: add new arrow for new menu

mjmonline:


src/components/icon/icons/thickArrowDown.jsx
```diff
@@ -1,3 +1,4 @@
+// DEPRECATED, Please use ChevronDownThick instead
```
JacobBlomgren: Now this svg is included twice, I would propose just re-exporting it in icons/index.js

PR #538, opened 2019-02-11
ci: only use node 10 and dont test build on publish

JacobBlomgren:
@iamstarkov and I talked, and since we don't ship any code that uses node, we don't actually need to run tests with node 8.

Also only runs build in the test stage for PRs, not on publish. That would be unnecessary because the publish step then proceeds to run build again for the publishing.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
+node_js:
+  - 10
 
 before_install:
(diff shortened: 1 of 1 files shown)


iamstarkov:
> @iamstarkov and I talked, and since we don't ship any code that uses node, we don't actually need to run tests with node 8.

affirmative

PR #541, opened 2019-02-13
chore: fix deprecated husky configs and cleanup package.json

JacobBlomgren:
I got some husky warnings on my machine, turns out that they have changed how it's supposed to be configured https://github.com/typicode/husky#upgrading-from-014 But thankfully they had a `./node_modules/.bin/husky-upgrade` so this is just a run of that and now committing and pushing works without warnings.

### DIFF
--- a/package.json
+++ b/package.json
@@ -51,20 +51,6 @@
       "git add"
     ]
   },
-  "nyc": {
-    "extension": [
-      ".js",
-      ".jsx"
-    ],
-    "all": true,
-    "instrument": false,
-    "sourceMap": false,
-    "reporter": [
-      "lcov",
-      "text-summary",
-      "text"
-    ]
-  },
   "author": "Gustaf Zetterlund <gustaf.zetterlund@nordnet.se>",
   "repository": "nordnet/nordnet-ui-kit",
   "bugs": {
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -137,5 +134,12 @@
     "sinon": "7.2.3",
     "validate-commit-msg": "2.14.0",
     "webpack": "4.29.3"
+  },
+  "husky": {
+    "hooks": {
+      "commit-msg": "validate-commit-msg",
+      "pre-commit": "lint-staged",
+      "pre-push": "run-p lint test"
```
jesperwiner: no pre-publish ?

JacobBlomgren: Nice detection! I don't think it's a git hook https://git-scm.com/docs/githooks however. @iamstarkov maybe you know more.

iamstarkov: `prepublish` as in `{pre,post}anyNpmScriptName` is a script to be invoked implicitly by npm right before it will invoke publish

iamstarkov: it is not a git hook

JacobBlomgren:
> why nyc config isnt needed anymore?
> at least it shouldnt be in this pr or change the title.

You're right. We don't use it anymore, as you can see it was removed here: https://github.com/nordnet/nordnet-ui-kit/pull/521/files but the config was forgotten behind.

PR #544, opened 2019-02-14
sameRow prop in radio group and some jest cleanup.

JacobBlomgren:
Primarily this adds a prop to let radio inputs be on the same row. 

Also for jest I found some minor things to fix after the migration from mocha, like jsdom being provided as an environment by default by jest, and an old istanbul babel plugin laying around that is unneeded (and uninstalled) now.


src/components/radio-group/radio-group.jsx
```diff
@@ -1,13 +1,21 @@
 import PropTypes from 'prop-types';
 import React from 'react';
+import injectSheet from 'react-jss';
+import cn from 'classnames';
+
+const styles = {
+  sameRow: {
```
jesperwiner: prob would had chosen to call it "layout" or "orientation" prop instead with like "horizontal" / "vertical" and default it to vertical

JacobBlomgren: Ok I'm on it

PR #557, opened 2019-02-19
chore: add codecov instead of coveralls

JacobBlomgren:
Well, coveralls was already disabled but we need a replacement ✨


iamstarkov:
markdown basics, so everybody are on the same page about md footnotes https://daringfireball.net/projects/markdown/basics

PR #567, opened 2019-02-25
fix: add defaults

jesperwiner:


JacobBlomgren:
Gonna close this as agreed to IRL since we basically ended up doing the same job twice in parallel.

PR #570, opened 2019-02-28
feat(editableinput): adds sensitive prop to EditableInput

anneback:
adds sensitive prop which will display the value as four asterisks if it is active

### DIFF
--- a/src/components/editable-input/editable-input.jsx
+++ b/src/components/editable-input/editable-input.jsx
@@ -7,6 +7,7 @@ import styles from './editable-input-styles';
 import { Button, Icon, Input, Spinner } from '../..';
 import HelpText from '../input/help-text';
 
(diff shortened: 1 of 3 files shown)


src/components/editable-input/editable-input.jsx
```diff
@@ -7,14 +7,15 @@ import styles from './editable-input-styles';
 import { Button, Icon, Input, Spinner } from '../..';
 import HelpText from '../input/help-text';
 
+const SENSITIVE_PLACEHOLDER_VALUE = '****';
```
KitsuneKelso: Just an idea that struck me: Maybe we should allow the user to pass in a value to have as the placeholder, but then just default back to this one. What do you think? :)

anneback: I added it

PR #574, opened 2019-03-06
chore: styleguidist 9

JacobBlomgren:
As you can see, the examples got a _lot_ clearer 🎉

### DIFF
--- a/styleguide.config.js
+++ b/styleguide.config.js
@@ -63,6 +63,9 @@ module.exports = {
   styleguideComponents: {
     Wrapper: path.join(__dirname, 'documentation', 'wrapper.jsx'),
   },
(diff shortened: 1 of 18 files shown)


styleguide.config.js
```diff
@@ -75,6 +78,9 @@ module.exports = {
     },
     devServer: {
       disableHostCheck: true,
+      headers: {
+        'Access-Control-Allow-Origin': '*',
```
JacobBlomgren: This is a solution from here: https://github.com/styleguidist/react-styleguidist/issues/1298#issuecomment-469951138 

It looks like styleguidist will eventually work without this, but I think it's fine for now.

PR #576, opened 2019-03-08
fix(typography): remove lineheight

jesperwiner:


JacobBlomgren:
Didn't we decide on doing something "smarter" for each size? Otherwise, good 👊

PR #579, opened 2019-03-11
fix: disallow extrabold for secondary, tertiary, caption

JacobBlomgren:


src/styles/typography.js
```diff
@@ -19,19 +19,20 @@ export default function createTypography() {
     ...ifSmallDevice(14, 16),
     fontWeight: WEIGHTS[weight] || WEIGHTS.regular,
   });
+  const disallowExtrabold = weight => (weight === 'extrabold' ? WEIGHTS.bold : WEIGHTS[weight]);
```
iamstarkov: coerce, is better

iamstarkov: perhaps, even create a separate weights hash object and just use it

JacobBlomgren: @iamstarkov I did that first, but then I went the function route because then we can change extrabold --> bold

PR #581, opened 2019-03-11
fix(input): adjust positioning and margins of help text

KitsuneKelso:
solves problem of help text bleeding into input and label


KitsuneKelso:
This is a solution for situations where something like this can happen. This is an exaggerated example, but proves the problem at hand.
image

KitsuneKelso:
If you can't check out the now-build for this, this is the behaviour of the proposed solution right now:
Screenshot 2019-03-11 at 17 40 30

PR #583, opened 2019-03-14
fix(editableinput): removes fixed size for editbutton on editableinput

anneback:


JacobBlomgren:
@anneback Since there was no description, might I just ask what the bug fix was? I tried to replicate how it was before, and the icon seems smaller with this change. It's definitely an OK PR and all, I'm just trying to gather what the purpose was?

KitsuneKelso:
> @anneback Since there was no description, might I just ask what the bug fix was? I tried to replicate how it was before, and the icon seems smaller with this change. It's definitely an OK PR and all, I'm just trying to gather what the purpose was?

@JacobBlomgren the problem was that the edit button was getting an oblong shape when the width and height were being set as they were. The other places where we use the edit button it's always styled without setting a width or height, so we needed this one to match them, as well as get rid of the non-perfect roundness.

This is what it looked like before the fix:

Screenshot 2019-03-15 at 08 52 46

Compare it to the button a bit further down in the image, that one is without any size adjustments.

JacobBlomgren:
@KitsuneKelso thank you, super clear description! good catch

PR #592, opened 2019-03-19
chore: fix commitlint config

JacobBlomgren:
Add [plugin](https://github.com/watson/commitlint-config-squash-pr) for allowing longer squashed messages (since a squash adds the PR number) and disable the annoying scope case rule so that for instance `feat(Button)` is allowed (same as our internal config).

### DIFF
--- a/package.json
+++ b/package.json
@@ -48,8 +48,10 @@
   },
   "commitlint": {
     "extends": [
+      "squash-pr",
       "@commitlint/config-conventional"
-    ]
+    ],
+    "rules": { "scope-case": [0] }
   },
   "lint-staged": {
     "*.{js,jsx}": [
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -48,8 +48,10 @@
   },
   "commitlint": {
     "extends": [
+      "squash-pr",
       "@commitlint/config-conventional"
-    ]
+    ],
+    "rules": { "scope-case": [0] }
```
iamstarkov: can you elaborate on this?

JacobBlomgren: It's from our internal config, so this is to allow for instance `feat(Button): ...` instead of only `feat(button): ...`, which you ran into the other day.

iamstarkov: nice

PR #608, opened 2019-04-04
feat: remove border-radius

jesperwiner:
adjusted some button paddings to make it look better


iamstarkov:
@jesperwiner does yarn install work fine to you?

PR #625, opened 2019-05-03
fix: update font sizes

mjmonline:
Updated font sizes according to design

### DIFF
--- a/src/styles/typography.js
+++ b/src/styles/typography.js
@@ -6,33 +6,42 @@ const WEIGHTS = {
   extrabold: 800,
 };
 
+const FONT_FAMILY = '"Nordnet Sans Mono", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif';
+
+const commonFontStyles = {
+  fontFamily: FONT_FAMILY,
+  letterSpacing: 'normal'
+}
+
 export default function createTypography() {
   const ifSmallDevice = (smallFontSize, largeFontSize) => ({
-    fontFamily:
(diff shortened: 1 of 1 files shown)


mjmonline:
Please don't merge this branch yet

src/styles/typography.js
```diff
@@ -17,16 +17,16 @@ export default function createTypography() {
     },
   });
   const primary = ({ weight = 'regular' } = {}) => ({
-    ...ifSmallDevice(14, 16),
+    ...ifSmallDevice(16, 16),
```
iamstarkov: does it make sense to have a function then? I mean it includes media query, so that will be included in all primary texts

iamstarkov: and perhaps will slow down the paint/draw/layout events

mjmonline: Good remarks, fixed

src/styles/typography.js
```diff
@@ -6,33 +6,42 @@ const WEIGHTS = {
...
+}
+
 export default function createTypography() {
   const ifSmallDevice = (smallFontSize, largeFontSize) => ({
-    fontFamily:
-      '"Nordnet Sans Mono", -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, sans-serif',
+    ...commonFontStyles,
     fontSize: smallFontSize,
-    letterSpacing: 'normal',
     [`@media only screen and (min-width: ${SMALL_DEVICE}px)`]: {
```
shmidt-i: SMALL_DEVICE is still `360` here, need to change to 480

mjmonline: @shmidt-i fixed here https://github.com/nordnet/nordnet-ui-kit/pull/632

PR #629, opened 2019-05-07
feat: a11y colors

JacobBlomgren:


src/styles/palette.js
```diff
@@ -1,11 +1,10 @@
-import color from './color';
-
-export default function createPalette({ type = 'light' }) {
-  const palette = { ...color };
+import createColor from './color';
 
+export default function createPalette({ type = 'light', a11yColors = false }) {
```
JacobBlomgren: don't be distracted by `type`, it's not used anywhere.

PR #631, opened 2019-05-10
More a11y color fixes

JacobBlomgren:
So I accidentally mixed up the a11y and normal colors with each other 😅Also I added more a11y alternatives for the brand colors.

### DIFF
--- a/src/styles/color.js
+++ b/src/styles/color.js
@@ -33,5 +33,5 @@ export default ({ a11yColors = false }) => ({
   cta: a11yColors ? '#0046FF' : '#2D67FF',
   positive: a11yColors ? '#00D200' : '#008A00',
   negative: a11yColors ? '#FF1900' : '#E81700',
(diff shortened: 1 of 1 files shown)


src/styles/color.js
```diff
@@ -30,8 +30,8 @@ export default ({ a11yColors = false }) => ({
 
   // FUNCTIONAL COLOURS FOR PRODUCT DEVELOPMENT
 
-  cta: a11yColors ? '#2D67FF' : '#0046FF',
-  positive: a11yColors ? '#008A00' : '#00D200',
-  negative: a11yColors ? '#E81700' : '#FF1900',
-  index: a11yColors ? '#FFCF00' : '#C15700',
+  cta: a11yColors ? '#0046FF' : '#2D67FF',
```
jesperwiner: think these still are wrong
0046FF, 00D200, FF1900 .   are normal colors .  not a11y colors

PR #652, opened 2019-10-14
fix(Spinner): use jss10 scoped keyframes

Dhq:
Trying to scope didn't work out great when jss added automatic scoping for keyframe ids :)


iamstarkov (approved):
very nice

